PR #1, opened 2021-05-11
Feature/test connectivity

peterbriggs42:
Initial scaffolding for Code42 Phantom app using py42 module. Only implements `test_connectivity` action at this point. Automated testing to come in later PRs.


code42_connector.py
```diff
@@ -0,0 +1,156 @@
...
+        action_id = self.get_action_identifier()
+
+        self.debug_print("action_id", self.get_action_identifier())
+
+        if action_id == 'test_connectivity':
+            ret_val = self._handle_test_connectivity(param)
+
+        return ret_val
+
+    def initialize(self):
```
alanag13: can we order these functions in a way that makes it easier to understand what's going on if you read this from top to bottom:

```
__init__
initialize
handle_action
_create_client
_handle_test_connectivity
# future action handlers go here
```

for example, when `_create_client` is right under `__init__`, it kind of looks like the client could get created with a `_cloud_instance` of `None`, even though that woudn't actually happen

code42_connector.py
```diff
@@ -0,0 +1,156 @@
...
+            return action_result.set_status(phantom.APP_ERROR, f"Could not create py42 client: {e}")
+
+        return phantom.APP_SUCCESS
+
+    def _handle_test_connectivity(self, param):
+        # Add an action result object to self (BaseConnector) to represent the action for this param
+        action_result = self.add_action_result(ActionResult(dict(param)))
+        self.save_progress("Connecting to endpoint")
+
+        if not self._create_client(action_result):
```
alanag13: would it make sense for us to call `_create_client` at the end of `initialize()` so that we won't have to include this little snippet in every single action handler?

alanag13: I'm assuming `initialize()` gets called automatically before every action through some kind of black magic because I don't see it being called anywhere in this code.

alanag13:
@peterbriggs42 did you check with Tony re: whether the logos used should be Code42 logos or Incydr logos?

code42_connector.py
alanag13: I think I figured this out, `main()` calls `_handle_action()` (which must be defined on `BaseConnector`), which must then call `initialize()` followed by `handle_action()`.

peterbriggs42: That's a good point. I modeled the client creation piece after the AWS apps I found [here](https://github.com/phantomcyber/phantom-apps/blob/next/Apps/phawscloudtrail/awscloudtrail_connector.py#L171), but I tried it the way you suggested and it appears to work just fine. I'm inclined towards calling it in `initialize()` like you suggest to reduce code duplication.

peterbriggs42:
> @peterbriggs42 did you check with Tony re: whether the logos used should be Code42 logos or Incydr logos?

(Not sure why I'm unable to respond to this issue directly) Yep I checked with Tony just now. He is following up with marketing folks to confirm. 

[Update] Tony confirmed that we want to be using the Code42 logos for the Phantom app.

alanag13 (approved):

PR #2, opened 2021-05-17
Add unit testing, documentation

peterbriggs42:


CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+```bash
+export PHANTOM_VM_IP_ADDR=0.0.0.0 # Replace with IP address for Phantom VM
+./util.sh move
+```
+
+Open the Phantom web app and login as the `admin` user. Navigate to `Apps > Unconfigured Apps` and find the Code42 App.
+Click Configure New Asset to supply the Console URL, username, and password to connect with. Fill out the fields in Asset Info
+and Asset Settings. Save the Asset Configuration and then click Test Connectivity to test your connection. 
+
+## Creating a virtual environment
```
alanag13: I think some of our other projects have instruction on how to create a venv on windows. Some of our QA folks would probably benefit from that.

exclude_files.txt
```diff
@@ -1,2 +1,11 @@
 .git*
-*.py
```
alanag13, line 2: I think you still want to exclude `*.py` files. The compile process turns these into `.pyc` files, and their corresponding source files should no longer be needed.

tests/test_connector.py
```diff
@@ -0,0 +1,66 @@
...
+@fixture
+def py42_client(mocker):
+    return mocker.MagicMock(spec=py42.sdk.SDKClient)
+
+
+@fixture
+def connector(request, py42_client):
+    connector = code42_connector.Code42Connector()
+    connector._client = py42_client
+    yield connector
```
alanag13: is there a reason this is `yield` and not `return`?

util.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+main() {
+  case $1 in
+  "")
+    echo "Kindly supply a command" && exit 0
+    ;;
+  move)
```
alanag13: another capability I can see being useful here would be copying the source dir from the vm to the local machine. I will often lazily use `vi` try out a change on the vm to see if it works.

`move` doesn't make it obvious which direction the files are going so especially if we had the above option, we'll want to come up with something a little less ambiguous (`deploy`?)

tests/test_connector.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def test_handle_connectivity(mocker, connector):
```
alanag13: These tests would probably be valuable:
- `from_local_account()` method gets called with the expected params
- `users.get_current()` gets called

exclude_files.txt
peterbriggs42: `exclude_files.txt` affects the compilation of the tarball on the host machine [here](https://github.com/phantomcyber/phcode42v2/blob/feature/tests/util.sh#L10); it's not actually consumed by the Phantom `compile_app.pyc` script. So since we need all the Python source files in the tarball I think this line ought to remain deleted.

.gitignore
```diff
@@ -0,0 +1,5 @@
+build
+dist
+phcode42v2.egg-info
+phantomstubs/phantomstubs.egg-info
+phcode42v2.tgz
```
antazoey: No newline at end of file

exclude_files.txt
alanag13: well _that's_ confusing -- why is this file provided by Phantom if nothing internal to Phantom consumes it? Is it perhaps only relevant to _some_ of the options of `compile_app.pyc`?

tests/test_connector.py
peterbriggs42: Think this was some copy+pasta from trying out the code generation feature of the new Phantom dev tooling. Removed.

code42_connector.py
```diff
@@ -64,8 +61,11 @@ def _handle_test_connectivity(self, param):
         action_result = self.add_action_result(ActionResult(dict(param)))
         self.save_progress("Connecting to endpoint")
 
-        response = self._client.users.get_current()
-        util.print_response(response)
+        try:
+            self._client.users.get_current()
+        except Exception as e:
+            exception_message = e.args[0].strip()
```
antazoey: What if an exception occurs here that has no args?

antazoey: Example: 

```
class MyErr(Exception):
"""Test"""

e = MyErr()

# e.args is an empty tuple
```

code42_connector.py
```diff
@@ -64,8 +61,11 @@ def _handle_test_connectivity(self, param):
         action_result = self.add_action_result(ActionResult(dict(param)))
         self.save_progress("Connecting to endpoint")
 
-        response = self._client.users.get_current()
-        util.print_response(response)
+        try:
+            self._client.users.get_current()
+        except Exception as e:
```
antazoey: Not sure what style we are supposed to follow, but PEP8 recommends avoiding single character variable names.

phantomstubs/phantom/action_result.py
```diff
@@ -0,0 +1,9 @@
+class ActionResult:
```
antazoey: I am guessing more is coming here regarding this?
A doc str would ease my newbie curiosity

phantomstubs/phantom/base_connector.py
```diff
@@ -0,0 +1,15 @@
...
+        pass
+
+    def save_progress(self, status):
+        pass
+
+    def get_action_identifier(self):
+        pass
+
+    def debug_print(self, name, value):
+        pass
```
antazoey: Missing newline at end of file

setup.py
```diff
@@ -0,0 +1,13 @@
...
+
+setup(
+    name="phcode42v2",
+    version="1.0.0",
+    description="Code42 for Phantom",
+    packages=find_packages(),
+    install_requires=["requests>=2.3",
+                      "pytest==4.4.0",
+                      "pytest-mock==1.10.3",
+                      "py42>=1.14.1",
```
alanag13: I think this will pull py42 (and phantomstubs, if something named that were to exist) from PyPI unless they already exist in the user's venv.

I think https://www.python.org/dev/peps/pep-0440/#direct-references can be used here to point to the wheel of py42 that we have in the repo, similar to this example: https://stackoverflow.com/a/65902365/12015509

phantomstubs/phantom/action_result.py
antazoey: Oh, are we to override this?
If that is the case, doc strings like `"""Override"""` are nice

setup.py
unparalleled-js, line 11: Is the tar file still needed if py42 is required here?

antazoey: Should this be `>=1.14.2` ?

setup.py
```diff
@@ -0,0 +1,13 @@
...
+                      "phantomstubs"]
```
alanag13: We could actually move all of `phantomstubs` to just be in a `phantom` folder under the main project, and just have the compile logic skip that dir.

alanag13: Something like that might look a little odd on the surface and what we're doing should probably be called out in the readme

tests/test_connector.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+logger = getLogger(name=__name__)
+
+
+NULL_VALUE = object()
+
+
+@fixture
+def py42_sdk(mocker):
```
antazoey: I tend to think it is a good idea to distinguish pytest fixtures that are mocks from ones that are actual class instances (such as `connector`) via prefixing the mocks with `mock_`.

tests/test_connector.py
```diff
@@ -0,0 +1,66 @@
...
+@fixture
+def connector(request, py42_client):
+    connector = code42_connector.Code42Connector()
+    connector._client = py42_client
+    return connector
+
+
+def test_handle_connectivity(mocker, connector):
+    param = {}
+    mocker.patch('phantom.base_connector.BaseConnector.add_action_result', return_value=ActionResult(dict(param)))
```
antazoey: Just a style thing, but the other files use double quotes for strings, which I like because they are easier to spot.

tests/test_connector.py
```diff
@@ -0,0 +1,66 @@
...
+    set_status = mocker.patch('phantom.action_result.ActionResult.set_status')
+
+    connector._handle_test_connectivity(param)
```
unparalleled-js, line 41: It is a little odd that we are testing a protected method.. Does `handle_test_connectivity()` need to be a protected method? I am guessing it is called outside the connector class, no?

tests/test_connector.py
```diff
@@ -0,0 +1,66 @@
...
+
+    set_status.assert_called_with(phantom.APP_SUCCESS)
+
+
+def test_handle_connectivity_sets_error_status_if_sdk_throws_exception(mocker, connector, py42_client):
+    param = {}
+    mocker.patch('phantom.base_connector.BaseConnector.add_action_result', return_value=ActionResult(dict(param)))
+    py42_client.users.get_current.side_effect = HTTPError(mock.Mock(status=401))
```
antazoey: This HTTPError, in this broadened state, is never possible in py42... Can we wrap this in a `Py42UnauthorizedError` to simulate a real user experience?

tests/test_connector.py
```diff
@@ -0,0 +1,66 @@
...
+    set_status = mocker.patch('phantom.action_result.ActionResult.set_status')
+
+    connector._handle_test_connectivity(param)
+
+    set_status.assert_called_with(phantom.APP_ERROR, mocker.ANY, mocker.ANY)
+
+
+def test_handle_action_calls_handle_connectivity(mocker, connector):
+    param = {}
+    handle_test_connectivity = mocker.patch('code42_connector.Code42Connector._handle_test_connectivity',
```
antazoey: In general, we should avoid mocking internal methods and instead treat them as implementation details.

Ideally, the tests above should probably be testing `handle_action` instead of `handle_test_connectivity` but still test the the same functionality.

At that point, I don't think this test (`test_handle_action_calls_handle_connectivity`) would even be needed

util.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+main() {
+  case $1 in
```
antazoey: Can we explicitly convert the first arg to a named variable?  Otherwise, errors may occur that are harder to understand, like `unbound variable $1`.

util.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
antazoey, line 1: Some documentation on the purpose of the script and possible commands here would be valuable.

util.sh
```diff
@@ -0,0 +1,21 @@
...
+  "")
+    echo "Kindly supply a command" && exit 0
+    ;;
+  deploy)
+    echo "Tarring the ball..."
+    pushd .. && tar -cvf phcode42v2/phcode42v2.tgz -X phcode42v2/exclude_files.txt phcode42v2/* && popd
+    echo "Moving to remote..."
+    sshpass -p "phantom" scp phcode42v2.tgz "phantom@${PHANTOM_VM_IP_ADDR}":/home/phantom
+    echo "Untarring on remote..."
+    sshpass -p "phantom" ssh phantom@10.0.0.41 "rm -rf phcode42v2 && tar -xvf phcode42v2.tgz && cd phcode42v2 && phenv python /opt/phantom/bin/compile_app.pyc -i"
```
antazoey: What IP address is this? Is this your local VM's?

util.sh
```diff
@@ -0,0 +1,21 @@
...
+    ;;
+  *)
+    echo "Not a valid command" && exit 0
+  esac
+}
+
+main "$@"
```
antazoey, line 45: Missing newline at end of file

util.sh
```diff
@@ -0,0 +1,21 @@
...
+  deploy)
+    echo "Tarring the ball..."
+    pushd .. && tar -cvf phcode42v2/phcode42v2.tgz -X phcode42v2/exclude_files.txt phcode42v2/* && popd
+    echo "Moving to remote..."
+    sshpass -p "phantom" scp phcode42v2.tgz "phantom@${PHANTOM_VM_IP_ADDR}":/home/phantom
+    echo "Untarring on remote..."
+    sshpass -p "phantom" ssh phantom@10.0.0.41 "rm -rf phcode42v2 && tar -xvf phcode42v2.tgz && cd phcode42v2 && phenv python /opt/phantom/bin/compile_app.pyc -i"
+    ;;
+  *)
+    echo "Not a valid command" && exit 0
```
antazoey: Should invalid commands exit with a code of `1` instead?

phantomstubs/phantom/action_result.py
alanag13: everything in the `phantomstubs` folder is just dummy implementations of stuff in the `phantom` namespace, because the `phantom` lib is only available inside of a running phantom instance. It's a cheap way to be able to mock it.

setup.py
antazoey: Why are we using the wheel instead of pypi?

alanag13: this is the only allowed way to include dependencies in a shipped phantom app (we can use PyPI when running tests since the tests won't actually be included in the shipped app).

tests/test_connector.py
alanag13: instead of patching, we could create a utility (note: not fixture) method that produces a version of the connector with a replaced `get_action_identifier`:

```python
def create_fake_connector(action_identifier):

    def fake_get_action_idenfifer():
        return action_identifier
        
    connector = code42_connector.Code42Connector()
    connector.get_action_identifier = fake_get_action_identifier
    return connector
```

Then we could do what Jules is suggesting by forcing tests to take the expected path:
```python
@fixture
def test_connectivity_connector(py42_client):
   connector = create_fake_connector("test_connectivity")
   connector._client = py42_client
   return connector
```
then `test_connectivity_connector.handle_action()` should automatically serve as a test that the correct internal method is called (whatever output or other effects we're testing would presumably fail if that _werent_ the case).

code42_connector.py
alanag13: they have some sort of linter built into their compile utility. I'm going to try to see if we can actually get it to run as part of CI.

CONTRIBUTING.md
```diff
@@ -0,0 +1,95 @@
...
+NOTE: If you are consistently getting timeouts connecting to the Phantom VM, try disconnecting from the VPN and re-attempting.
+
+
+## Deploying the app
+
+Run the below command to install the Code42 app on Phantom. 
+
+```bash
+export PHANTOM_VM_IP_ADDR=0.0.0.0 # Replace with IP address for Phantom VM
+./util.sh deploy
```
antazoey, line 26: I got `sshpass` not found... That must be a requirement I need to install. Should we mention that somewhere?

antazoey: Took a minute to find decent installation instructions, but this one worked: https://stackoverflow.com/a/62623099/667301

util.sh
antazoey: Should this be `PHANTOM_VM_IP_ADDR`?

CONTRIBUTING.md
antazoey: After installing `sshpass` and using `PHANTOM_VM_IP_ADDR`, I get the following errors:

```
Unauthorized access prohibited.
 ____ ____ ____ ____ ____ ____ ____ ____ ____ ____ ____ ____ ____ ____
||s |||p |||l |||u |||n |||k |||> |||P |||h |||a |||n |||t |||o |||m ||
||__|||__|||__|||__|||__|||__|||__|||__|||__|||__|||__|||__|||__|||__||
|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|
Permission denied, please try again.
Untarring on remote...
Unauthorized access prohibited.
 ____ ____ ____ ____ ____ ____ ____ ____ ____ ____ ____ ____ ____ ____
||s |||p |||l |||u |||n |||k |||> |||P |||h |||a |||n |||t |||o |||m ||
||__|||__|||__|||__|||__|||__|||__|||__|||__|||__|||__|||__|||__|||__||
|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|/__\|
Permission denied, please try again
```

I am trying to figure out what, do I need to use `sudo` or something?

util.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
```
antazoey, line 2: I think we need a `set -e` at the top here.

Currently, I am seeing failures when trying to move the tar to the remote but then still tries to untar anyways.  The script / command should fail if one part of the command fails.

CONTRIBUTING.md
antazoey: I realize now that it is because my password is not `phantom`... Did I miss this instruction?

Should the password also be an environment variable?

CONTRIBUTING.md
```diff
@@ -0,0 +1,95 @@
...
+to create a Phantom account. Phantom will approve your registration and send you credentials for https://my.phantom.us.
+
+Next, log in and go [here](https://my.phantom.us/downloads/) to download the Phantom VM image.
+
+Follow these [instructions](https://docs.splunk.com/Documentation/Phantom/4.10.3/Install/InstallOVA) to install Splunk Phantom 
+as a VM image. You do NOT need to complete the section labeled "Configure the network settings for the virtual machine".
+
+Use these [default credentials](https://docs.splunk.com/Documentation/Phantom/4.10.3/Install/Reference) to log in.
+Confirm that you can connect to the VM via SSH on your host machine. You can use `hostname -I` on the remote machine to look up your IP address.
+Also confirm that you can access the Phantom web app at `https://{phantom VM IP address}`.
```
antazoey: The spaces in `phantom VM IP address` cause a weird underline thing in VSCode, could we put dashes or something in between the words?

antazoey:
I am bringing this up because I anticipate QA will if we don't:

The error message that occurs when you input a bad username / password is like a whole exception message and looks kind of intimidating. Do you know if there is a way to make this simpler? Like catch the 401 and just say "Invalid Credentials" or something?

I am seeing the error in Phantom when configuring a new Asset and inputting bad creds


Edit: Yeah Nevermind, I see the custom message portion... Maybe some newlines between that and the exception message, if possible, would help, otherwise disregard


Edit 2:  Actually, I see the custom exception in the code but the one I get in the UI does not seem to be the same..?
Here is what I see in the UI:

```
Exception Occurred. Failure in HTTP call 401 Client Error: Unauthorized for url: ....
```

But in the code, I see logic for printing 

```
"Unable to connect to Code42."
```

unparalleled-js:
> I am bringing this up because I anticipate QA will if we don't:
> 
> The error message that occurs when you input a bad username / password is like a whole exception message and looks kind of intimidating. Do you know if there is a way to make this simpler? Like catch the 401 and just say "Invalid Credentials" or something?
> 
> I am seeing the error in Phantom when configuring a new Asset and inputting bad creds

Yeah Nevermind, I see the custom message portion... Maybe some newlines between that and the exception message, if possible, would help, otherwise disregard

antazoey:
`__main__` calls a method `_handle_action()` in the `Code42Connector` in file `code42_connector.py` but I don't see that method, I only see the method `handle_action()`.  Is it supposed to call `handle_action()`? Could that be why I am not seeing the expected error in my earlier comment?

alanag13:
> `__main__` calls a method `_handle_action()` in the `Code42Connector` in file `code42_connector.py` but I don't see that method, I only see the method `handle_action()`. Is it supposed to call `handle_action()`?

`_handle_action` is defined on the Base class and calls `initialize()` followed by `handle_action()`.

Peter and I discussed and agreed that we should effectively treat all of `main()` as if it were a third party dependency since:

- it was generated
- we have no intention of ever changing it
- mocking it all is non-trivial.

 That's why all of our tests set things up as if `initialize()` has already been called.

>  Could that be why I am not seeing the expected error in my earlier comment?

I think this is because the error occurs during `initialize()` and not during the handler (`from_local_account()` actually makes a call in order to validate credentials).

phantomstubs/phantom/action_result.py
peterbriggs42: I'll add a note on this either to Contributing.md or as a docstring

CONTRIBUTING.md
peterbriggs42: Yep my mistake, realized when I was helping Teri that `sshpass` is an undocumented dependency, as well as that I accidentally hardcoded the password in there. Should be fixed now

util.sh
```diff
@@ -9,9 +9,17 @@ main() {
     echo "Tarring the ball..."
     pushd .. && tar -cvf phcode42v2/phcode42v2.tgz -X phcode42v2/exclude_files.txt phcode42v2/* && popd
     echo "Moving to remote..."
-    sshpass -p "phantom" scp phcode42v2.tgz "phantom@${PHANTOM_VM_IP_ADDR}":/home/phantom
+    scp phcode42v2.tgz "phantom@${PHANTOM_VM_IP_ADDR}":/home/phantom
     echo "Untarring on remote..."
-    sshpass -p "phantom" ssh phantom@10.0.0.41 "rm -rf phcode42v2 && tar -xvf phcode42v2.tgz && cd phcode42v2 && phenv python /opt/phantom/bin/compile_app.pyc -i"
+    ssh "phantom@${PHANTOM_VM_IP_ADDR}" "rm -rf phcode42v2 && tar -xvf phcode42v2.tgz && cd phcode42v2 && phenv python /opt/phantom/bin/compile_app.pyc -i"
+    ;;
+  deploy-bypass)
```
alanag13, line 27: do we need both of these deploy options? It seems like the second one is just better (though we should document the need for `sshpass` somewhere).

If we _do_ want both, there's enough in common between the two commands here that some stuff should probably be combined into shared functions

exclude_files.txt
peterbriggs42: I think I actually just misinterpreted what Dan said in his presentation. Now I believe that the `exclude_files.txt` is only included in the app wizard boilerplate as a convenience to suggest to the user "if you want to exclude files from the app deployed to the Phantom server, use this file to omit those files during archive creation". Tough to say for certain without access to source, but I tested how the `compile_app.pyc` responds to presence / absence of `exclude_files.txt` and I did not observe its presence having any effect.

alanag13: It looks like that script has a `-g` flag to perhaps pass this file to?

```
[phantom@phantom ~]$ phenv python3 /opt/phantom/bin/compile_app.pyc -h
usage: compile_app.pyc [-h] (-s SINGLE_PYFILE | -i | -t | -c) [-a APP_DIR]
                       [-d] [-e] [-g IGNORE_FILE]

optional arguments:
  -h, --help            show this help message and exit
  -s SINGLE_PYFILE, --single_pyfile SINGLE_PYFILE
                        Compile a Single python file and exit
  -i, --install_app     Install app after compilation
  -t, --create_tarball  Only create the app tarball and exit, compilation or
                        installation is not carried out
  -c, --compile_app     Compile the app and exit
  -a APP_DIR, --app_dir APP_DIR
                        app directory
  -d, --exclude_flake   Dont run flake
  -e, --continue_on_error
                        Stop on error
  -g IGNORE_FILE, --ignore_file IGNORE_FILE
                        files that contains the list of files to ignore, by
                        default it is .compile_app.ignore
```

util.sh
peterbriggs42: sshpass isn't supported at all for Windows so I added `deploy` to provide support there. Wasn't able to find a solution that is supported for Windows and macOS/Linux. I'll be sure to add a note for the sshpass dependency.

CONTRIBUTING.md
peterbriggs42: Creating a virtual environment is only necessary if you want to run unit tests -- I'll add a note to make that more explicit.

setup.py
peterbriggs42: Ahh nice I spent a bit of time trying to get it to use the wheel without success. This direct references functionality is just what I was looking for!

setup.py
peterbriggs42: I think it's a bit more readable to have the top level directory for the stubs read `phantomstubs` rather than just `phantom`. The `phantomstubs` dir is already excluded from the archive compilation in `exclude_files.txt` -- I'm kind of confused by what you are suggesting here?

alanag13: I'm basically suggesting that we fool the test environment into thinking that `phantom` is just part of our project and not an external dependency. Instead of having the stubs be installed from a folder outside the source, we can just include it in the project directly, but ensure that it doesn't actually ship with the code we distribute.

code42_connector.py
peterbriggs42: Ahh good catch. Fixed

setup.py
antazoey: I like what @alanag13 is suggesting. At minimum, it would be nice to have the package name be just `phantom` so the imports work

tests/test_connector.py
peterbriggs42: Implemented Alan's suggestion

peterbriggs42:
> > `__main__` calls a method `_handle_action()` in the `Code42Connector` in file `code42_connector.py` but I don't see that method, I only see the method `handle_action()`. Is it supposed to call `handle_action()`?
> 
> `_handle_action` is defined on the Base class and calls `initialize()` followed by `handle_action()`.
> 
> Peter and I discussed and agreed that we should effectively treat all of `main()` as if it were a third party dependency since:
> 
>     * it was generated
> 
>     * we have no intention of ever changing it
> 
>     * mocking it all is non-trivial.
> 
> 
> That's why all of our tests set things up as if `initialize()` has already been called.
> 
> > Could that be why I am not seeing the expected error in my earlier comment?
> 
> I think this is because the error occurs during `initialize()` and not during the handler (`from_local_account()` actually makes a call in order to validate credentials).

Alan and I discussed this. `initialize` is the preferable location to instantiate the py42 client; however, it poses a problem because it doesn't have access to the `action_result` object used for reporting error message to the console. It might be possible to work around this by applying a decorator to each action method that instantiates the py42 client, but we didn't want to hold up this PR by figuring it out.

For now, I moved the client instantiation back into the body of the action method. We'll need to copy+paste this instantiation in other actions we implement for the time being. Also worth noting that this is how all the AWS Phantom apps handle instantiating the boto3 client, so it's at least precedented.

setup.py
peterbriggs42: Implemented Alan's suggestion

exclude_files.txt
peterbriggs42: Alan and I discussed this. For the `deploy` command, we decided to perform the exclusion at tar creation time. If in the future we extend this script to add a command just for compiling the app, we will use the `-g` argument of the `compile_app.pyc` script against the same `exclude_files.txt` file.

alanag13 (approved):
There are some minor things in the tests that I think we can improve (I think patches are being a little overused in places where mocks or an update to the stub itself could suffice), but right now I thinks it's most valuable to get this through so that others can benefit from it.

PR #3, opened 2021-05-20
Add/Remove DE + other small changes 

antazoey:
* Adds two new actions, Add Departing Employee and Remove Departing Employee
* Other small adjustments


code42_connector.py
```diff
@@ -28,6 +31,19 @@ def __init__(self):
...
         self._password = None
         self._client = None
+        self._action_map = {
+            self.TEST_CONNECTIVITY_ACTION_ID: lambda x: self._handle_test_connectivity(x),
+            self.ADD_DEPARTING_EMPLOYEE_ACTION_ID: lambda x: self._handle_add_departing_employee(x),
+            self.REMOVE_DEPARTING_EMPLOYEE_ACTION_ID: lambda x: self._handle_remove_departing_employee(x)
+        }
+
+    @property
+    def client(self):
```
antazoey, line 41: A lazily loaded client makes it so we don't have to initialize it in every action, now that it doesnt happen at the initialize connector stage

code42_connector.py
```diff
@@ -136,14 +173,16 @@ def main():
...
 
         if session_id is not None:
-            in_json['user_session_token'] = session_id
-            connector._set_csrf_info(csrftoken, headers['Referer'])
+            in_json["user_session_token"] = session_id
+            if csrftoken and headers:
+                connector._set_csrf_info(csrftoken, headers["Referer"])
 
-        ret_val = connector._handle_action(json.dumps(in_json), None)
+        json_string = json.dumps(in_json)
```
antazoey, line 182: Every change in this part is just making strings use double quotes besides this one,
I separated out this variable to attempt at making the debugging experience better

code42_connector.py
alanag13: I'm pretty sure that every action just calls `main()` and will end up creating a new Connector, so a new client will always end up getting created on every action.

tests/test_connector.py
```diff
@@ -1,100 +1,145 @@
...
 
 @fixture
 def test_connectivity_connector(mock_py42_client):
-    connector = create_fake_connector("test_connectivity")
+    connector = create_fake_connector(Code42Connector.TEST_CONNECTIVITY_ACTION_ID)
     connector._client = mock_py42_client
     return connector
 
 
 class TestCode42Connector(object):
```
alanag13, line 24: I think it probably makes sense to split the tests up by action category, otherwise this will become a monster

code42_connector.py
```diff
@@ -28,6 +31,19 @@ def __init__(self):
         self._username = None
         self._password = None
         self._client = None
+        self._action_map = {
```
alanag13, line 34: I'm doing something similar to this in my ci/cd branch im working on :-)

antazoey: It's in XSOAR too, overlap deja vu

tests/test_connector.py
antazoey: Should they be in different files or just different classes?
If different files, we can move the fixtures to a conftest and that will even more-so lighten up their monstrosities

tests/test_connector.py
```diff
@@ -1,100 +1,145 @@
...
+        self, mocker, mock_py42_client, mock_result_adder
+    ):
+        param = {"username": "test@example.com", "departure_date": "2030-01-01"}
+        result = ActionResult(dict(param))
+        result.update_summary = mocker.MagicMock()
+        mock_result_adder.return_value = result
+        http_response = mocker.MagicMock(spec=Response)
+        http_response.text = json.dumps({"users": [{"userUid": "TEST_USER_UID"}]})
+        mock_py42_client.users.get_by_username.return_value = Py42Response(http_response)
+        connector = create_fake_connector(Code42Connector.ADD_DEPARTING_EMPLOYEE_ACTION_ID)
```
alanag13: I generally tend to avoid referencing constants that are in the project when writing tests. The main reason for this is because readers gain a better understanding of how the system is supposed to work when the input is not hidden behind a var that is defined in a separate project. For example, when I read this PR, I do not know what the id for this action actually is, but I would if it just appeared as a raw string here.

antazoey: I usually agree, and can put them like that if you want... My concern is that there are just so many raw strings here, it is maddening to know what is what.

tests/test_connector.py
alanag13: yeah that's exactly what I was thinking, different files

tests/test_connector.py
```diff
@@ -1,100 +1,145 @@
...
 
         set_status.assert_called_with(phantom.APP_SUCCESS)
 
-    def test_handle_test_connectivity_calls_users_get_current(self, mocker, mock_py42_client, connector):
+    def test_handle_test_connectivity_calls_users_get_current(self, connector, mock_result_adder):
         param = {}
-        mocker.patch("py42.sdk.from_local_account", return_value=mock_py42_client)
-        mocker.patch("phantom.base_connector.BaseConnector.add_action_result", return_value=ActionResult(dict(param)))
-
+        mock_result_adder.return_value = ActionResult(dict(param))
```
alanag13, line 36: from what I can tell looking at existing projects, this is literally always the value that is passed to `add_action_result`.

code42_connector.py
```diff
@@ -65,19 +74,45 @@ def _handle_test_connectivity(self, param):
...
         return action_result.set_status(phantom.APP_SUCCESS)
 
+    def _handle_add_departing_employee(self, param):
+        self.save_progress("In action handler for: {0}".format(self.get_action_identifier()))
+        action_result = self.add_action_result(ActionResult(dict(param)))
+        username = param["username"]
+        departure_date = param.get("departure_date")
+        user_id = self._get_user_id(username)
+        self.client.detectionlists.departing_employee.add(user_id, departure_date=departure_date)
+        action_result.update_summary({})
```
alanag13: I think this and/or `set_summary` are basically the equivalents of the table outputs in XSOAR. It might make sense to choose a couple relevant properties for each of these actions and add them here

tests/test_connector.py
antazoey: What is? `param`?  `param` will be different per action context I think

code42_connector.py
antazoey: Like the parameters?
I don't think there are responses for these API calls

code42_connector.py
```diff
@@ -65,19 +74,45 @@ def _handle_test_connectivity(self, param):
...
+        status_message = "{} was added to the departing employee list".format(username)
```
alanag13: we can use f-strings here :-)

antazoey: Can we use them everywhere then? I was copying the techniques I had already observed in the app, so I will change them everywhere if that is the case

code42_connector.py
antazoey: Yes, but there are helper methods that may also need to use the client, such as the method for getting the user ID.. So rather than assuming the SDK is initialized at the start of every method, this make it so that it will just happen the first time it is used.

tests/test_connector.py
antazoey: They have been swapped out 👍🏻

code42_connector.py
antazoey: Oh, and I don't think the logic in `__main__` is anything except for debugging actions, from what I can tell.

In a Playbook setting, I am not sure if the client actually gets re-initialized per action.

antazoey:
I am seeing an odd behavior with /slash commands and optional arguments in the Analyst view (not in a Playbook).
I have an action with both a required and an optional argument.  When using the Action UI (the button that says > Action in the top right), I can specify both args and the action gets called once and everything is expected / working.  However, when I call the same action using the CLI, it seems to execute the action twice, once per argument... The second time fails because it is missing the required argument, and the first time does not include the optional argument.

My CLI command looks like this:
`/action command_name "AppName" required_arg --opt_arg_flag opt_arg`

When generating test JSON from the CLI versus the ActionUI, they are clearly different... The CLI has two context blocks in the argument list, one per arg, while the Action UI contains both args in the same context and therefore the example code works for accessing required and optional args

tl;dr:

I can't specify optional args using slash commands but I can using the UI and am not sure what is going there and am trying to figure it out

Edit:  I have mentioned this in the slack community

code42_connector.py
alanag13: > In a Playbook setting, I am not sure if the client actually gets re-initialized per action.

If it's not through `main()`, then I have no idea how Phantom could possibly discover the `Code42Connector` class and use it. I'd be surprised

I get what you're saying about helper methods though. I have a different way of doing this in my branch that I think should address that concern, but there's nothing wrong with keeping it this way for now.

code42_connector.py
alanag13: no, but you could still do something like output the ID and username of the user that was added / removed / etc. Isn't that what we do in XSOAR?

alanag13: There is also `add_data` which I think is basically an actual return value that could be consumed by further actions. I think `summary` is just for display purposes.

tests/test_connector.py
alanag13: it will be yes, sorry -- this wasn't really feedback, more just a comment that I noticed that everywhere in the actual code `ActionResult(dict(param))` is _always_ what is passed to `add_action_result`, but yes, the value of `param` will be different depending on the action.

code42_connector.py
```diff
@@ -65,19 +74,45 @@ def _handle_test_connectivity(self, param):
...
-            self._client = py42.sdk.from_local_account(self._cloud_instance, self._username, self._password)
-            self._client.users.get_current()
+            self.client.users.get_current()
         except Exception as exception:
             return action_result.set_status(phantom.APP_ERROR, f"Unable to connect to Code42: {str(exception)}")
 
         self.save_progress("Test Connectivity Passed")
         return action_result.set_status(phantom.APP_SUCCESS)
 
+    def _handle_add_departing_employee(self, param):
```
alanag13, line 84: Should we handle specific exceptions where that we can give nice error messages, like when you try to add a user that's already added, user doesn't exist, etc?

antazoey: I thought we would have to, but the nice error messages from py42 are already custom enough imo when they happen in the UI

antazoey: e.g. 

```
handle_action exception occurred. Error string: 'User with ID 921286907298179098 is already on the departing-employee list.'
```

alanag13: nice!

code42_connector.py
antazoey: Oh okay I see, yeah let me take a closer look at this and probably copy XSOAR

Edit: also, I have no output-related tests... I would think `add_data()` would be important tests.. I usually don't worry too much about outputting data for POST like actions (contain / response versus investigate), but it never hurts.  I have also heard people recommending adding tests for table output too.

code42_connector.py
alanag13: everywhere in this project, yes

code42_connector.py
alanag13: `add_data` would definitely be important to test since if that were to change it could break Playbooks and such.
Since `summary` is just visual that is probably less critical but probably still nice to do

code42_connector.py
antazoey: The reason I think this is because in this article: https://docs.splunk.com/Documentation/Phantom/4.10.3/DevelopApps/Tutorial

The section titled `Debugging` is the only place from what I can tell where `main()` is mentioned.. Also, notice how it imports `pudb` at the top of the method. There is no reason a debugger module would be needed in production.

Here is quote from the article: `Running the ipinfoio_connector.py script under a debugger runs the app in standalone mode. In other words, for the most part, it does not talk with the Splunk Phantom platform. Any progress messages that the app sends are dumped on the console and not sent to the platform. This type of execution helps in debugging the app code that interacts with the external device, in this case the ipinfo.io servers.`

That sounds like typically connectors are not run in standalone mode.

I could be wrong, but just sharing why I think this.

code42_connector.py
antazoey: It does seem like we might get a lot of this for free, like the action's parameters used seem to automatically go to outputs under a sub-obj `parameters`, as well as the status of the action etc.

Edit: also this: https://github.com/phantomcyber/phcode42v2/pull/3#issuecomment-845403937

antazoey:
I notice a lot of other projects accept different values per argument, like a single argument can be an ID or a username, etc... The reason for this is because if you see an ID or a username anywhere in the app, in notes or outputs or evidence etc, you can right click it and select actions to do it on.  So, you may see a UserId on a particular alert and can `add_to_departing_employee`.. Then, you might see a username from a 3rd party app, such as HRM, and you would have the same option, `add_to_departing_employee`.... It is the same action acting on different artifact-types. I normally would not be that into that, but from the UX of someone clicking around, investigating in Phantom, I appreciate the flexibility... Otherwise, they'd have to first figure out the correct username / ID (whatever the action requires) and would lose the power-user ability

tests/test_detectionlists_actions.py
```diff
@@ -0,0 +1,44 @@
+import json
+
+from py42.response import Py42Response
+from requests import Response
+
+from phantom.action_result import ActionResult
+from tests.conftest import create_fake_connector
+
+
+class TestCode42DetectionListsConnector(object):
```
alanag13, line 28: might be good to test that `handle_action` returns success / failure under the appropriate circumstances

code42_connector.py
timabrmsn: Could we do this and not have to have any maps of identifiers to methods? 

```
ret_val = getattr(self, "_handle_{0}".format(action_id))(param)
```

alanag13: you could, but that would create a strict dependency on what handler functions now _have_ to be named. It sometimes isn't a big deal, but I would generally discourage this kind of coupling.

code42_connector.py
```diff
@@ -65,19 +74,47 @@ def _handle_test_connectivity(self, param):
...
+        self.save_progress(f"In action handler for: {self.get_action_identifier()}")
+        action_result = self.add_action_result(ActionResult(dict(param)))
+        username = param["username"]
+        departure_date = param.get("departure_date")
+        user_id = self._get_user_id(username)
+        response = self.client.detectionlists.departing_employee.add(user_id, departure_date=departure_date)
+        action_result.add_data(response.data)
+        action_result.update_summary({"userId": user_id})
```
alanag13: we know the username from what the user passed in too, so that might be nice to add here.

code42_connector.py
```diff
@@ -65,19 +74,47 @@ def _handle_test_connectivity(self, param):
...
         return action_result.set_status(phantom.APP_SUCCESS)
 
+    def _handle_add_departing_employee(self, param):
+        self.save_progress(f"In action handler for: {self.get_action_identifier()}")
+        action_result = self.add_action_result(ActionResult(dict(param)))
+        username = param["username"]
+        departure_date = param.get("departure_date")
+        user_id = self._get_user_id(username)
+        response = self.client.detectionlists.departing_employee.add(user_id, departure_date=departure_date)
+        action_result.add_data(response.data)
```
alanag13, line 91: what does `response.data` hold for these APIs? Is it worth including?

antazoey: See the tests! (for exact content)

There is nothing in the remove call, but the add call has stuff.. Adding the response by default here doesn't really hurt and seems to be the norm. I think we may still have to define outputs in the json file in order for them to work in playbooks.

antazoey: So, if like XSOAR, there are 3 groups:

1.) The raw full response data
2.) The table data
3.) Action outputs (for playbook action chaining)

code42_connector.py
antazoey: Thanks. Also noticed that the casing for these should be in snake_case

alanag13 (approved):

PR #5, opened 2021-05-21
Fix/outputs-for-playbooks-and-table

antazoey:
I found issues with outputs when testing these actions in a Playbook. Good news though, playbooks are great and with these changes, outputs now work for these actions.

### DIFF
--- a/code42.json
+++ b/code42.json
@@ -89,7 +89,7 @@
                     "contains": [
                         "email"
                     ],
-                    "column_name": "username",
+                    "column_name": "Username",
                     "column_order": 0
                 },
                 {
@@ -173,19 +173,19 @@
                     "contains": [
                         "email"
                     ],
-                    "column_name": "username",
+                    "column_name": "Username",
                     "column_order": 0
                 },
                 {
                     "data_path": "action_result.parameter.departure_date",
                     "data_type": "string",
-                    "column_name": "departure_date",
+                    "column_name": "Departure Date",
                     "column_order": 1
                 },
                 {
                     "data_path": "action_result.parameter.note",
                     "data_type": "string",
-                    "column_name": "note",
+                    "column_name": "Note",
                     "column_order": 2
                 },
                 {
--- a/code42.json
+++ b/code42.json
@@ -112,16 +112,7 @@
                 },
                 {
                     "column_name": "UserId",
-                    "data_path": "action_result.data.*.userId",
-                    "data_type": "string",
-                    "example_values": [
-                        "1"
-                    ],
-                    "column_order": 1
-                },
-                {
-                    "column_name": "Username",
-                    "data_path": "action_result.data.*.userName",
+                    "data_path": "action_result.data.*.user_id",
                     "data_type": "string",
                     "example_values": [
                         "1"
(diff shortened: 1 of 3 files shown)


code42.json
```diff
@@ -111,31 +111,14 @@
                     "data_type": "numeric"
                 },
                 {
-                    "column_name": "UserId",
+                    "column_name": "User ID ",
```
antazoey, line 114: Column names can and usually have spaces and look pretty

code42.json
```diff
@@ -111,31 +111,14 @@
...
                     "data_path": "action_result.data.*.userId",
                     "data_type": "string",
                     "example_values": [
                         "1"
                     ],
                     "column_order": 1
                 },
-                {
-                    "column_name": "Username",
```
antazoey, line 123: Username is already includes via `action_result.parameter` and there is no username on the remove response anyway.

code42.json
```diff
@@ -111,31 +111,14 @@
...
-                    "data_path": "action_result.data.*.userName",
-                    "data_type": "string",
-                    "example_values": [
-                        "1"
-                    ],
-                    "column_order": 1
-                },
-                {
-                    "column_name": "CreatedAt",
```
antazoey, line 132: There is no createdAt on the `remove` response

code42.json
```diff
@@ -239,10 +222,13 @@
                     "example_values": [
                         "1"
                     ],
+                    "contains": [
```
antazoey, line 225: Plugs in a lot easier if uses the same contains fields as other actions may expect.
There is also a `user name` contains type but it seemed buggy, makes it hard to input raw strings for some reason, requires data to be classified...

code42_connector.py
```diff
@@ -103,8 +103,8 @@ def _handle_remove_departing_employee(self, param):
         action_result = self._add_action_result(param)
         username = param["username"]
         user_id = self._get_user_id(username)
-        response = self.client.detectionlists.departing_employee.remove(user_id)
-        action_result.add_data(response.data)
+        self.client.detectionlists.departing_employee.remove(user_id)
+        action_result.add_data({"userId": user_id})
```
antazoey, line 107: In the JSON file, we were saying that we were outputting the ID when we were not in fact.. Also, remember the username is already included in output and can be used in subsequent actions in a playbook since it was a parameter and that happens by default.

unparalleled-js:
Hmm... Now I am noticing something weird with Table outputs outside of playbooks and am investigating that

peterbriggs42 (approved):

code42.json
```diff
@@ -190,19 +173,19 @@
...
                     ],
-                    "column_name": "username",
+                    "column_name": "Username",
                     "column_order": 0
                 },
                 {
                     "data_path": "action_result.parameter.departure_date",
                     "data_type": "string",
-                    "column_name": "departure_date",
+                    "column_name": "Departure Date",
```
antazoey, line 182: These seem to be all caps anyway, so the most important part is replacing underscores with spaces...

It's just a table output though.

Also, @peterbriggs42 pointed out that the table output seem to be driven off the output rather than the summary

PR #6, opened 2021-05-24
Feature/len actions 2

antazoey:
New actions:

* Add high risk employee
* Remove high risk employee
* List departing employees
* List high risk employees

I have tested both with the Action menu as well as in Playbooks

### DIFF
--- a/code42.json
+++ b/code42.json
@@ -185,6 +185,123 @@
                 "type": "table"
             },
             "versions": "EQ(*)"
+        },
+        {
+            "action": "remove high risk employee",
+            "identifier": "remove_high_risk_employee",
+            "description": "Remove a high risk employee from the high risk employees detection list",
+            "verbose": "",
+            "type": "contain",
+            "read_only": false,
+            "parameters": {
+                "username": {
+                    "description": "Username of the employee to remove",
+                    "data_type": "string",
+                    "required": true,
+                    "primary": true,
+                    "contains": [
+                        "email"
+                    ],
+                    "value_list": [],
+                    "default": "",
+                    "order": 0
+                }
+            },
+            "output": [
+                {
+                    "data_path": "action_result.parameter.username",
+                    "data_type": "string",
+                    "contains": [
+                        "email"
+                    ],
+                    "column_name": "username",
+                    "column_order": 0
+                },
+                {
+                    "data_path": "action_result.status",
+                    "data_type": "string",
+                    "column_name": "status",
+                    "column_order": 1
+                },
+                {
+                    "data_path": "action_result.message",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "summary.total_objects",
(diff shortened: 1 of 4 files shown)


code42.json
```diff
@@ -61,11 +61,129 @@
             "output": [],
             "versions": "EQ(*)"
         },
+        {
+            "action": "add departing employee",
```
antazoey, line 65: Diff looks a little funny because I moved the `add` commands above the `remove`

code42.json
```diff
@@ -61,11 +61,129 @@
...
+            "identifier": "add_departing_employee",
+            "description": "Add a departing employee to the departing employees detection list",
+            "verbose": "There is a short delay before you can query the updated results.",
+            "undo": "remove departing employee",
```
antazoey, line 69: I added these `undo` sections - they are neat. They make it so you can right click on old action and there is now an "undo" action available which just guides you through the opposite action

code42_connector.py
```diff
@@ -106,7 +116,63 @@ def _handle_remove_departing_employee(self, param):
...
+        status_message = f"{username} was removed from the high risk employees list"
+        return action_result.set_status(phantom.APP_SUCCESS, status_message)
+
+    def _handle_list_high_risk_employees(self, param):
+        self._log_action_handler()
+        action_result = self._add_action_result(param)
+        filter_type = param.get("filter_type", HighRiskEmployeeFilters.OPEN)
+        results_generator = self.client.detectionlists.high_risk_employee.get_all(filter_type=filter_type)
+        total_count = None
+        for page in results_generator:
```
alanag13, line 163: Normally to handle `total_count` I would do something like:

```python
page = None
for page in results_generator:
    employees = page.data.get("items", [])
    for employee in employees: 
         action_result.add_data(employee)
 
total_count = page.data.get("totalCount", 0) if page else None
```
`page` will be set to the last page if there was at least one page of data.

code42.json
timabrmsn: that's neat!

code42_connector.py
alanag13: you could do this for departing employees too. To me this makes it more clear that we aren't actually setting the total count on each iteration.

alanag13 (approved):

antazoey:
I have been working with Peter because we have been confused by the summary portion... It seems like the ascii tables are not driven from the summary but rather some little text in the activity list, same place where the `status_message` is... However, when including both a `status_message` and a `summary`, only the `status_message appears` there.  List commands seem to often display the total count and not a `status_message`, so I remove the `status_message` from both list commands.  I have seen other commands from other apps use both the summary and the `status_message`.. The summary data is available in the context...

antazoey:
@alanag13 I made the change discussed via not using status message for list commands and not using summary for add / remove, still approved?

PR #7, opened 2021-05-24
Chore/ci cd

alanag13:
- add github actions workflow for unit testing
 - separate src packages from tests to fix IDE test debugging
 - enhance stub classes to have realistic behaviors in order to reduce the need for mocks
 - refactor tests
 - fix tarballing to handle exclusions properly
 
 The `CONTRIBUTING` guide also highlights a cool new capability: you can now either run the tests in your local dev environment as normal _or_, if the repo is on a Phantom server, you can run the tests in such a way that the _real_ phantom SDK classes are used instead of the stubs.


util.sh
```diff
@@ -8,13 +8,13 @@
...
 #   ssh: SSH into your Phantom VM.
-#   open-web: Open your default web browser at `https://<phantom-vm-ip:9999`.
+#   open-web: Open your default web browser at `https://<phantom-vm-ip:9999>`.
 
 set -eo pipefail
 
 make_tar() {
   echo "Tarring the ball..."
-  pushd .. && tar -cvf phcode42v2/phcode42v2.tgz -X phcode42v2/exclude_files.txt phcode42v2/* && popd
+  tar -cvf phcode42v2.tgz -X ./exclude_files.txt -H .
```
alanag13: I found that this command was not properly excluding the files because when we use `pushd` the lines in `exclude_files` no longer match anything.

tests/conftest.py
```diff
@@ -41,3 +36,36 @@ def create_mock_response(mocker, response_data):
     response = mocker.MagicMock(spec=Response)
     response.text = json.dumps(response_data)
     return Py42Response(response)
+
+
+def assert_success(connector):
```
alanag13, line 41: These are some utility functions that make writing the tests a bit more concise. One thing I tried to always do is have each test check the status was successful (or error, if that was what it was supposed to be)

src/phantom/action_result.py
```diff
@@ -0,0 +1,52 @@
+from phantom.app import APP_ERROR
```
alanag13: By adding a very small amount of simple logic to these stubs, we can make them behave like the real thing and eliminate a lot of the need for mocks. All these methods exist on these classes in the real world.

src/code42_connector.py
```diff
@@ -0,0 +1,238 @@
...
+        self._state = self.load_state()
+
+        config = self.get_config()
+        self._cloud_instance = config["cloud_instance"]
+        self._username = config["username"]
+        self._password = config["password"]
+
+        return phantom.APP_SUCCESS
+
+    def handle_action(self, param):
```
alanag13: I made `handle_action` take care of:
- creating the client
- creating and adding the action result (which, in addition to `param` is passed to all actions now)
- logging which action is about to be handled

This removed a fair amount of boilerplate code from the actions.

setup.py
```diff
@@ -8,7 +8,8 @@
     name="phcode42v2",
     version="1.0.0",
     description="Code42 for Phantom",
-    packages=find_packages(),
+    packages=find_packages("src"),
```
alanag13: This is what ultimately should fix the issue with some IDEs not being able to debug through tests. Basically, without this, `pip install` causes the tests to get installed too, which causes weirdness.

tests/test_connector.py
```diff
@@ -22,46 +15,17 @@ def test_connectivity_connector(mock_py42_client):
 
 
 class TestCode42Connector(object):
```
alanag13, line 16: A lot of the tests got significantly simplified, notably due to no longer needing `mock_result_adder`.
Another thing I tried to do was have every test assert that the success status code was set (or the failure one, if that's what was needed). Because of that, we were able to get rid of tests that only checked the status code and nothing else.

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "python.testing.pytestArgs": [
+        "tests"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.nosetestsEnabled": false,
+    "python.testing.pytestEnabled": true
+}
```
antazoey: No newline at end of file

code42_connector.py
```diff
@@ -38,29 +38,19 @@ def __init__(self):
...
         self._client = None
         self._action_map = {
-            self.TEST_CONNECTIVITY_ACTION_ID: lambda x: self._handle_test_connectivity(x),
-            self.ADD_DEPARTING_EMPLOYEE_ACTION_ID: lambda x: self._handle_add_departing_employee(x),
-            self.REMOVE_DEPARTING_EMPLOYEE_ACTION_ID: lambda x: self._handle_remove_departing_employee(x),
-            self.LIST_DEPARTING_EMPLOYEES_ACTION_ID: lambda x: self._handle_list_departing_employees(x),
-            self.ADD_HIGH_RISK_EMPLOYEE_ACTION_ID: lambda x: self._handle_add_high_risk_employee(x),
-            self.REMOVE_HIGH_RISK_EMPLOYEE_ACTION_ID: lambda x: self._handle_remove_high_risk_employee(x),
-            self.LIST_HIGH_RISK_EMPLOYEES_ACTION_ID: lambda x: self._handle_list_high_risk_employees(x)
+            self.TEST_CONNECTIVITY_ACTION_ID: self._handle_test_connectivity,
```
antazoey: This causes Pycharm to get mad when calling the `action_handler`, claiming unexpected arguments

antazoey:
Are you able to deploy this? When I try, I get `bash: line 0: cd: phcode42v2: No such file or directory
`

code42_connector.py
```diff
@@ -133,31 +120,27 @@ def _handle_list_departing_employees(self, param):
         action_result.update_summary({"total_count": total_count})
         return action_result.set_status(phantom.APP_SUCCESS)
 
-    def _handle_add_high_risk_employee(self, param):
-        self._log_action_handler()
-        action_result = self._add_action_result(param)
+
+    # HIGH RISK EMPLOYEE ACTIONS
```
antazoey: I am pretty sure the linting they do will not be ok with style here..
If you switch this comment to be a string-comment like `""" HIGH RISK EMPLOYEE ACTIONS """` and put a newline before and after it, it will work

stubs/phantom/action_result.py
```diff
@@ -0,0 +1,52 @@
+from phantom.app import APP_ERROR
```
antazoey, line 1: This file uses a mix of tabs and spaces

code42_connector.py
```diff
@@ -71,57 +61,54 @@ def initialize(self):
...
+        if not action_handler:
+            return action_result.set_status(phantom.APP_ERROR, f"Code42: Action {action_id} does not exist.")
 
+        try:
+            if not self._client:
+                self._client = py42.sdk.from_local_account(self._cloud_instance, self._username, self._password)
+            self.save_progress(f"Code42: handling action {action_id}...")
+            return action_handler(param, action_result)
+        except Exception as ex:
+            msg = f"Code42: Failed execution of action {action_id}: {str(ex)}"
```
antazoey: I thought interpolating strings does `str(item)` under the head, regarding `{str(ex)}`

```python
>>> err = Exception("TestErr")
>>> test1 = f"this is an err: {err}"
>>> test2 = f"this is an err: {str(err)}"
>>> test1
'this is an err: TestErr'
>>> test2
'this is an err: TestErr'
```

code42_connector.py
```diff
@@ -71,57 +61,54 @@ def initialize(self):
...
+    def _handle_test_connectivity(self, param, action_result):
+        self._client.users.get_current()
         self.save_progress("Test Connectivity Passed")
         return action_result.set_status(phantom.APP_SUCCESS)
 
-    def _handle_add_departing_employee(self, param):
-        self._log_action_handler()
-        action_result = self._add_action_result(param)
+
+    # DEPARTING EMPLOYEE ACTIONS
```
antazoey: ` """ DEPARTING EMPLOYEE ACTIONS """ ` with new lines above and below, to prevent phlint errors

code42_connector.py
```diff
@@ -71,57 +61,54 @@ def initialize(self):
...
-            # Fix this with a decorator or something similar.
-            self.client.users.get_current()
-        except Exception as exception:
-            return action_result.set_status(phantom.APP_ERROR, f"Unable to connect to Code42: {str(exception)}")
+        if not action_handler:
+            return action_result.set_status(phantom.APP_ERROR, f"Code42: Action {action_id} does not exist.")
 
+        try:
+            if not self._client:
+                self._client = py42.sdk.from_local_account(self._cloud_instance, self._username, self._password)
```
antazoey, line 68: I think this still has the same problem as before, no?
When calling the `test_connectivity`, if your creds are false, it will fail before reaching the `test_connectivity` action handler,

alanag13:
> Are you able to deploy this? When I try, I get `bash: line 0: cd: phcode42v2: No such file or directory `

I busted something here, my bad, will fix

code42_connector.py
```diff
@@ -133,31 +120,27 @@ def _handle_list_departing_employees(self, param):
         action_result.update_summary({"total_count": total_count})
         return action_result.set_status(phantom.APP_SUCCESS)
 
-    def _handle_add_high_risk_employee(self, param):
-        self._log_action_handler()
-        action_result = self._add_action_result(param)
+
+    # HIGH RISK EMPLOYEE ACTIONS
+    def _handle_add_high_risk_employee(self, param, action_result):
```
antazoey, line 125: I can't see why this would not be ok, but just noticing that I don't really see any other app have any other parameters in their action-handler methods besides the single `param`.

setup.py
```diff
@@ -3,14 +3,18 @@
 from setuptools import find_packages, setup
 
 path_to_py42_wheel = os.path.join(os.getcwd(), 'wheels', 'py42-1.14.2-py2.py3-none-any.whl')
+path_to_stubs = os.path.join(os.getcwd(), 'stubs#egg=phantom')
 
 setup(
-    name="phcode42v2",
+    name="code42_connector",
```
antazoey: Seems kind of generic without `ph` somewhere in there

tests/conftest.py
```diff
@@ -41,3 +36,40 @@ def create_mock_response(mocker, response_data):
...
+    assert summary == expected_summary
+    assert status == phantom.app.APP_SUCCESS
+
+def assert_successful_message(connector, expected_message):
+    action_results = connector.get_action_results()
+    assert len(action_results) == 1
+    msg = action_results[0].get_message()
+    status = action_results[0].get_status()
+    assert msg == expected_message
+    assert status == phantom.app.APP_SUCCESS
```
antazoey: no newline at end of file

tests/test_detectionlists_actions.py
```diff
@@ -340,125 +294,86 @@ def test_handle_action_when_add_high_risk_employee_adds_response_to_data(
...
-        result.add_data = add_data_mock
-        mock_result_adder.return_value = result
         connector = _create_remove_hr_connector(mock_py42_with_user)
         connector.handle_action(param)
-        add_data_mock.assert_called_once_with({"userId": _TEST_USER_UID})
+        assert_successful_single_data(connector, {"userId": _TEST_USER_UID})
 
-    def test_handle_action_when_remove_high_risk_employee_and_is_successful_sets_success_status(
-        self, mocker, mock_py42_with_user, mock_result_adder
+    def test_handle_action_when_remove_departing_employee_and_is_successful_sets_success_status(
```
antazoey: This test is apparently a duplicate test? It is also defined on line 220

setup.py
```diff
@@ -3,14 +3,18 @@
 from setuptools import find_packages, setup
```
peterbriggs42: `import find_packages` is no longer needed here

tests/test_detectionlists_actions.py
```diff
@@ -340,125 +294,86 @@ def test_handle_action_when_add_high_risk_employee_adds_response_to_data(
...
-        result.set_status = set_status_mock
-        mock_result_adder.return_value = result
         connector = _create_list_hr_connector(mock_py42_with_high_risk_employees)
         connector.handle_action({})
-        set_status_mock.assert_called_once_with(1)
+        action_results = connector.get_action_results()
+        assert len(action_results) == 1
+        data = action_results[0].get_data()
+        assert data[0] == _MOCK_LIST_HIGH_RISK_EMPLOYEES_RESPONSE["items"][0]
+        assert data[1] == _MOCK_LIST_HIGH_RISK_EMPLOYEES_RESPONSE["items"][1]
```
antazoey: No new line at end of file

stubs/phantom/action_result.py
```diff
@@ -0,0 +1,52 @@
...
+
+class ActionResult:
```
peterbriggs42, line 3: Is this the actual decompiled source for this file? Or a stub you wrote with a mostly-functional implementation that matches the behavior of the actual class?

util.sh
```diff
@@ -8,13 +8,13 @@
...
 #   ssh: SSH into your Phantom VM.
-#   open-web: Open your default web browser at `https://<phantom-vm-ip:9999`.
+#   open-web: Open your default web browser at `https://<phantom-vm-ip:9999>`.
 
 set -eo pipefail
 
 make_tar() {
   echo "Tarring the ball..."
-  pushd .. && tar -cvf phcode42v2/phcode42v2.tgz -X phcode42v2/exclude_files.txt phcode42v2/* && popd
+  tar -cvf phcode42v2.tgz -X ./exclude_files.txt .
```
antazoey, line 17: This is the reason the deploy is failing for me

antazoey:
All phantom validation errors:

```
./code42_connector.py:86:5: E303 too many blank lines (2)
./code42_connector.py:87:5: E301 expected 1 blank line, found 0
./code42_connector.py:124:5: E303 too many blank lines (2)
./code42_connector.py:125:5: E301 expected 1 blank line, found 0
./code42_connector.py:156:5: E303 too many blank lines (2)
./code42_connector.py:167:1: E302 expected 2 blank lines, found 1
```

code42_connector.py
```diff
@@ -71,57 +61,54 @@ def initialize(self):
...
+        self._client.users.get_current()
         self.save_progress("Test Connectivity Passed")
         return action_result.set_status(phantom.APP_SUCCESS)
 
-    def _handle_add_departing_employee(self, param):
-        self._log_action_handler()
-        action_result = self._add_action_result(param)
+
+    # DEPARTING EMPLOYEE ACTIONS
+    def _handle_add_departing_employee(self, param, action_result):
```
peterbriggs42, line 84: Much prefer this signature with the injected `action_result`.

tests/test_detectionlists_actions.py
antazoey: Probably should be `high_risk_employee` instead of `departing` in the test sig.

tests/test_detectionlists_actions.py
```diff
@@ -1,8 +1,6 @@
 from pytest import fixture
 
-from phantom.action_result import ActionResult
-from tests.conftest import create_fake_connector
-from tests.conftest import create_mock_response
+from .conftest import assert_success, create_fake_connector, create_mock_response, assert_successful_single_data, assert_successful_message, assert_succesful_summary
```
antazoey: Why switch to relative import here?
Also could we at least use the tuple syntax here so this isn't going off the screen?

tests/test_detectionlists_actions.py
```diff
@@ -1,8 +1,6 @@
 from pytest import fixture
```
antazoey, line 1: That tests that end in `_sets_success_status` should probably be changed to `_sets_success_message()` since that seems to be more the focus of the tests now.

code42_connector.py
timabrmsn: What about just having the action map be a decorator on each function instead of enumerating them all?:

```
ACTION_MAP = {}

def action_key(key):
    def wrapper(f):
        ACTION_MAP[key] = f
        return f

    return wrapper
```
Less stuff to write and it ties the key closely to the method itself: 

```
    @action_key("test_connectivity")
    def _handle_test_connectivity(self, param):
```

stubs/setup.py
```diff
@@ -0,0 +1,11 @@
+import os
```
antazoey: `import os` is not needed

stubs/setup.py
```diff
@@ -0,0 +1,11 @@
...
+
+from setuptools import find_packages, setup
+
+setup(
+    name="stubs",
+    version="1.0.0",
+    description="Phantom Stubs for Code42 for Phantom",
+    packages=find_packages(),
+    install_requires=["requests>=2.3"]
```
antazoey: I noticed that `requests` on the server us up to version `2.24.0`.
Are we sure this is the minimum supported version?

antazoey: Wait... how come the stubs even need requests?

setup.py
```diff
@@ -3,14 +3,18 @@
...
 path_to_py42_wheel = os.path.join(os.getcwd(), 'wheels', 'py42-1.14.2-py2.py3-none-any.whl')
+path_to_stubs = os.path.join(os.getcwd(), 'stubs#egg=phantom')
 
 setup(
-    name="phcode42v2",
+    name="code42_connector",
     version="1.0.0",
     description="Code42 for Phantom",
-    packages=find_packages(),
-    install_requires=["requests>=2.3",
```
antazoey, line 12: How come we are removing `requests`?  It is imported in `code42_connector.py`

antazoey: Also (from other comment but means more here):

I noticed that requests on the server us up to version 2.24.0.
Should we make that is minimally the version we use?

alanag13: This was a mistake, `requests` should be here and not in the stubs

stubs/setup.py
alanag13: This was already in place, and was likely the result of it being copied from one of our other projects, we should have it be the same as on the server

stubs/phantom/action_result.py
alanag13: The latter, though I did determine this using decompilers

code42_connector.py
alanag13: I _realllllllly_ like this idea, it makes the association between the key and the function it's tied to super clear

tests/test_detectionlists_actions.py
alanag13: I don't think it has to, I think I did this while I was trying to get all the packages properly separated and never came back here.

code42_connector.py
antazoey: I like this way this looks!

util.sh
alanag13: should be fixed now, a long with the validation errors

code42_connector.py
alanag13: we can obviously change it without much trouble at all if we're ever presented with a reason we have to.

stubs/phantom/action_result.py
alanag13: remind me to never copy/paste anything ever again

code42_connector.py
alanag13: yes, but we will catch the exception, and the action result will be set to be an error. The caller wouldn't really care whether or not we actually reached the handler.

code42_connector.py
alanag13: mad as in squiggly lines? Hmm, vscode doesn't complain at all and tests definitely show this should be fine, is there a way we could make pycharm ignore this?

tests/test_detectionlists_actions.py
alanag13: it has the incorrect name, should be hr instead of de

code42_connector.py
antazoey: What does the error message look like though?
the original concern was that it was a giant blob.
I will check.

alanag13:
@unparalleled-js @timabrmsn @peterbriggs42 I think I've responded to all feedback here. I might open another PR to add some additional things to ci that might be nice to have but for now I'd like other people's PRs to be able to take advantage of this.

code42_connector.py
antazoey: Looks great! Better than it did before I think

code42_connector.py
antazoey: Yeah, I did, just thought it was weird as other apps did this way and pycharm didn't complain, made me think we had something fishy going on, but I didn't see anything.

I think the decorator pattern made this a non-issue now anyway

timabrmsn (approved):

stubs/phantom/action_result.py
antazoey: VSCode has some quick actions for converting tabs to spaces (and visa versa) as well as trimming lines. When I use vs code, I run those quick actions on everything

antazoey (approved):

stubs/setup.py
```diff
@@ -0,0 +1,8 @@
+from setuptools import find_packages, setup
```
peterbriggs42, line 1: Missing newline at EOF

peterbriggs42 (approved):

PR #8, opened 2021-05-25
Feature/user actions

timabrmsn:


code42_connector.py
```diff
@@ -61,11 +61,15 @@ def handle_action(self, param):
         action_result = self.add_action_result(ActionResult(dict(param)))
 
         if not action_handler:
-            return action_result.set_status(phantom.APP_ERROR, f"Code42: Action {action_id} does not exist.")
+            return action_result.set_status(
+                phantom.APP_ERROR, f"Code42: Action {action_id} does not exist."
+            )
```
timabrmsn: These are just `black` doing its thing

code42.json
```diff
@@ -629,6 +629,242 @@
...
+                    "primary": true,
+                    "contains": [
+                        "email"
+                    ],
+                    "value_list": [],
+                    "default": "",
+                    "order": 0
+                }
+            },
+            "output": [
```
antazoey, line 1656: Probably should have some sort of summary mentioned in the output, even if it's just the success message.
I believe the success message counts as the summary if you include this:

```
, 
{
  "data_path": "action_result.summary",
  "data_type": "string"
}

Same with `unblock` and any others
```

code42_connector.py
```diff
@@ -154,16 +164,83 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+            username=username,
+            email=username,
+            password=password,
+            first_name=param.get("first_name"),
+            last_name=param.get("last_name"),
+            notes=param.get("notes"),
+        )
+        user_id = response["userUid"]
+        action_result.add_data(response.data)
+        action_result.update_summary({"user_id": user_id, "username": username})
```
antazoey: Something we noticed:  If you include a summary and a status message, only the status message shows up in the UI, but if you leave off the status message, the summary shows up. There may be other places where it's visible but I have yet to find it.

tests/test_user_actions.py
```diff
@@ -0,0 +1,252 @@
...
+        connector = create_fake_connector("create_user", mock_py42_with_create_user)
+        connector.handle_action(param)
+        assert_successful_single_data(
+            connector, {"userUid": _TEST_USER_UID, "userId": _TEST_USER_LEGACY_ID},
+        )
+
+    def test_handle_action_when_creating_user_without_password_param_generates_password(
+        self, mocker, mock_py42_with_create_user
+    ):
+        mock_passgen = mocker.patch("code42_connector._generate_password")
```
antazoey: Instead of this, you can test that the generated password contains a capital letter, a lower case letter, is of a certain length, a symbol, and a number. That way, we are at least testing that it meets the criteria without caring about its actual value

code42.json
```diff
@@ -629,6 +629,242 @@
                 "type": "table"
             },
             "versions": "EQ(*)"
+        },
+                {
+            "action": "create user",
+            "identifier": "create_user",
+            "description": "Create a new Code42 user account",
```
antazoey: Do you think it's worth including a `"verbose":` mentioning that once a user is created, it may take a few minutes before you are able to add them to detection lists?

Tony did say there were use cases around that

antazoey:
Heads up that you have some phlint warning:

```
create user
      Missing render dictionary
      Parameter "org_uid" has invalid contains, "org_id". Contains must only contain alphanumeric characters and spaces, should be lower case.
      Following required data paths not in output list
        action_result.parameter.org_uid
        action_result.parameter.password
        action_result.parameter.first_name
        action_result.parameter.last_name
        action_result.parameter.notes
    block user
      Missing render dictionary
      Following required data paths not in output list
        action_result.data
        action_result.summary
        action_result.status
        action_result.message
    deactivate user
      Missing render dictionary
      Following required data paths not in output list
        action_result.data
        action_result.summary
        action_result.status
        action_result.message
    reactivate user
      Missing render dictionary
      Following required data paths not in output list
        action_result.data
        action_result.summary
        action_result.status
        action_result.message
    unblock user
      Missing render dictionary
      Following required data paths not in output list
        action_result.data
        action_result.summary
        action_result.status
        action_result.message

```

code42.json
```diff
@@ -629,6 +629,242 @@
...
+            "type": "contain",
+            "read_only": false,
+            "parameters": {
+                "org_uid": {
+                    "description":  "Org UID to create user in",
+                    "data_type": "string",
+                    "required": true,
+                    "primary": true,
+                    "contains": [
+                        "org_id"
```
antazoey: I am guessing that needs to be `org id` instead

code42.json
timabrmsn: yeah, makes sense

code42_connector.py
```diff
@@ -236,5 +314,16 @@ def main():
     exit(0)
 
 
+def _generate_password():
```
alanag13: Something about this being visible on an open source repo doesn't sit right with me. Maybe we should force users to supply a password, I'm going to talk to Tony.

antazoey: Why not just make password optional?  If not provided, leave the user in invited mode.

timabrmsn: Mainly because invited mode can be pretty confusing unless you actually go through the invite process. The user account is created, but doesn't show up in the normal org user list unless you click on the 'invited' tab. But I suppose that's just a problem in the product that we don't necessarily have to try to fix ourselves on this end.

code42.json
```diff
@@ -629,6 +629,242 @@
...
+                    "primary": true,
+                    "contains": [
+                        "email"
+                    ],
+                    "default": "",
+                    "order": 1
+                },
+                "password": {
+                    "description":  "Password for newly created user (defaults to randomly generated password, requiring reset before login)",
+                    "data_type": "string",
```
peterbriggs42: Should the data type for this parameter be `password`? I believe that setting it as such will do things like encrypt the values on Phantom's backend and obfuscate the values when entered in the UI

code42_connector.py
alanag13: it's worth noting that the invite behavior is what we do in XSOAR currently (I believe). I did reach out to Tony about this and it's all fine by him but he recommended we reach out to a PAL. Personally I think using the invite pattern is probably fine if properly documented.

antazoey: We could add a note to the `verbose` saying that by not including a password, the user will be in `invite` mode.

timabrmsn: Sounds good, I doubt phantom will actually be used much for creating users either way.

code42.json
timabrmsn: I don't think anything produces an org id contains type so I'll just remove it for now..

code42.json
```diff
@@ -1065,6 +1065,243 @@
...
+                    "column_name": "status",
+                    "column_order": 2
+                },
+                {
+                    "data_path": "action_result.message",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "action_result.summary",
```
antazoey: The summary is more accessible if you reference the props you add to it via dot notation,

like
```
 "data_path": "action_result.summary.user_id"
```

etc, and then you can remove this bare summary one

code42.json
```diff
@@ -1065,6 +1065,243 @@
...
+                "username": {
+                    "description":  "Username to create",
+                    "data_type": "string",
+                    "required": true,
+                    "primary": true,
+                    "contains": [
+                        "email"
+                    ],
+                    "default": "",
+                    "order": 1
```
antazoey, line 1423: The order skips `"order": 2`

code42.json
```diff
@@ -1065,6 +1065,243 @@
...
+                    "data_path": "action_result.parameter.username",
+                    "data_type": "string",
+                    "contains": [
+                        "email"
+                    ],
+                    "column_name": "Username",
+                    "column_order": 0
+                },
+                {
+                    "data_path": "action_result.data.userUid",
```
antazoey: I am little concerned we will confuse people by using `userUid` to mean the same as `userId`.

XSOAR does output the `userUid` as `UserId` for these user-actions.. However, XSOAR also does not output anything else for these methods.

Thoughts?

alanag13: I think we have other stuff that already uses `userId` too, so I think sticking with that makes sense if we intend to output it. Do we even have to output it though? All the actions take username as input 🤷🏼‍♂️

code42.json
```diff
@@ -1065,6 +1065,280 @@
...
+                    "data_type": "string"
+                }
+            ],
+            "versions": "EQ(*)"
+        },
+        {
+            "action": "block user",
+            "identifier": "block_user",
+            "description": "Blocks a user from accessing their Code42 account",
+            "verbose": "Blocks a user from accessing their Code42 account (endpoint backup/security processes continue while blocked)",
```
antazoey: Verbose comments need to end in periods or validation gets mad
I noticed this when I made a change to an app in `phantom-apps`

code42.json
```diff
@@ -1399,6 +1399,7 @@
             "action": "create user",
             "identifier": "create_user",
             "description": "Create a new Code42 user account",
+            "verbose":"There is a short delay before you will be able to add the new user to departing employee or high risk employee lits.",
```
antazoey: `lits` should be `list`

(also missing space between key and value)

antazoey (approved):
Yay. Just a spelling error

PR #9, opened 2021-05-26
Feature/lens 3

antazoey:
New actions:

* get_departing_employee
* get_highrisk_employee
* add_highrisk_tag
* remove_highrisk_tag


code42_connector.py
```diff
@@ -119,6 +130,20 @@ def _handle_list_departing_employees(self, param, action_result):
...
 
+    @action_handler_for("get_departing_employee")
+    def _handle_get_departing_employee(self, param, action_result):
+        username = param["username"]
+        user_id = self._get_user_id(username)
+
+        try:
+            response = self._client.detectionlists.departing_employee.get(user_id)
+            action_result.add_data(response.data)
+            action_result.update_summary({"is_departing_employee": True})
```
antazoey, line 141: The reason these `get` actions were requests in XSOAR is because they wanted to be able to check if a user was in fact a certain detection list occupant.

So, the summary tells them that and even provided a bool to trip off of.

If they are a list occupant, they will also get the data of the API response, such as the `userId` etc

code42_connector.py
```diff
@@ -154,6 +181,48 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+        return action_result.set_status(phantom.APP_SUCCESS)
+
+    @action_handler_for("add_highrisk_tag")
+    def _handle_add_high_risk_tag(self, param, action_result):
+        username = param["username"]
+        risk_tag = param["risk_tag"]
+        user_id = self._get_user_id(username)
+        response = self._client.detectionlists.add_user_risk_tags(user_id, risk_tag)
+        all_tags = response.data.get("riskFactors", [])
+        response["riskFactors"] = _convert_to_obj_list(all_tags, "tag")
```
antazoey, line 207: `riskFactors` gets converted to a list of objects all with a single key `tag`.
This makes it so that you can use `....*.tag` in playbooks and perform actions off every tag in the list

code42_connector.py
```diff
@@ -154,6 +181,48 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+    @action_handler_for("remove_highrisk_tag")
+    def _handle_remove_high_risk_tag(self, param, action_result):
+        username = param["username"]
+        risk_tag = param["risk_tag"]
+        user_id = self._get_user_id(username)
+        response = self._client.detectionlists.remove_user_risk_tags(user_id, risk_tag)
+        all_tags = response.data.get("riskFactors", [])
+        response["riskFactors"] = _convert_to_obj_list(all_tags, "tag")
+        action_result.add_data(response.data)
+        message = (
```
antazoey, line 221: The status message for `add` is just the risk tags from the response and a nice message. I tried both a summary and status message and ultimately felt the status message looked prettier in this case and the risk tags are available in the data anyway

The `remove` command additionally will output that the `user has not tags` if the tags are empty. The tags list should never be empty in the case of `add`

code42.json
```diff
@@ -629,6 +734,331 @@
...
+                    "order": 0
+                },
+                "risk_tag": {
+                    "description": "A risk tag to add to the user",
+                    "data_type": "string",
+                    "required": true,
+                    "primary": false,
+                    "default": "",
+                    "value_list": [
+                        "HIGH_IMPACT_EMPLOYEE",
```
antazoey, line 877: This is what creates the drop down in the UI

code42.json
```diff
@@ -629,6 +734,331 @@
...
+                        "1"
+                    ],
+                    "contains": [
+                        "email"
+                    ],
+                    "column_order": 4
+                },
+                {
+                    "column_name": "Risk Tag",
+                    "data_path": "action_result.data.*.riskFactors.*.tag",
```
antazoey, line 945: When uses as an input to another action, will do action on every tag in list

tests/test_detectionlists_actions.py
```diff
@@ -377,3 +510,164 @@ def test_handle_action_when_list_high_risk_employee_adds_response_items_to_data(
...
+    ):
+        param = {"username": "test@example.com"}
+        mock_py42_with_user.detectionlists.high_risk_employee.get.side_effect = (
+            Py42NotFoundError(mock.Mock(status=404))
+        )
+        connector = _create_get_hr_connector(mock_py42_with_user)
+        connector.handle_action(param)
+        assert_succesful_summary(connector, {"is_high_risk_employee": False})
+
+    def test_handle_action_when_add_high_risk_tag_calls_add_with_expected_args(
```
alanag13, line 546: can you add an `assert_success` to this test?

code42_connector.py
```diff
@@ -31,8 +31,11 @@ def wrapper(f):
     return wrapper
 
 
-class Code42Connector(BaseConnector):
+def _convert_to_obj_list(scalar_list, sub_object_key="item"):
```
alanag13, line 34: maybe the second arg should be required for this method. I'm not sure `items` occurs frequently enough for it to be considered a default for this, and making it required would force someone to think more closely about it and make it less likely for them to erroneously use `items`.

alanag13: nvm I misunderstood what this method does, this makes sense now

alanag13 (approved):
Found one very. minor thing with a test missing a success assertion, otherwise LGTM

tests/test_detectionlists_actions.py
antazoey: Yes, found a couple

PR #10, opened 2021-05-26
Chore/more ci cid

alanag13:
- convert `util.sh` to makefile
- add style checks for black, BOM, end of file, and trailing whitespace
- integrate phantom validation/tarring scripts into local and ci/cd builds


.gitignore
```diff
@@ -7,6 +7,7 @@ phcode42v2.tgz
 __pycache__/
 *.py[cod]
 *$py.class
+!build-scripts/**/*.pyc
```
alanag13, line 10: we need to include these `.pyc` files for the phantom build scripts to work

code42_connector.py
```diff
@@ -85,7 +87,9 @@ def _handle_add_departing_employee(self, param, action_result):
         username = param["username"]
         departure_date = param.get("departure_date")
         user_id = self._get_user_id(username)
-        response = self._client.detectionlists.departing_employee.add(user_id, departure_date=departure_date)
+        response = self._client.detectionlists.departing_employee.add(
```
alanag13, line 90: this is all just `black`

exclude_files.txt
```diff
@@ -1,17 +1,20 @@
 .git*
-.pytest_cache/
-*.egg-info/
-.vscode/
+*.pyc
```
alanag13, line 2: I changed the tarball command to be the script provided by phantom instead of `tar`.  The name `exclude_files.txt` is hard-coded into phantom as the name of the file to read from when looking for exclusions. It also doesn't seem to treat slashes the way we would expect, instead any folder or file that matches the pattern is what gets ignored

setup.py
```diff
@@ -2,20 +2,28 @@
 
 from setuptools import setup
 
-path_to_py42_wheel = os.path.join(os.getcwd(), 'wheels', 'py42-1.14.2-py2.py3-none-any.whl')
-path_to_stubs = os.path.join(os.getcwd(), 'stubs#egg=phantom')
+path_to_py42_wheel = os.path.join(
+    os.getcwd(), "wheels", "py42-1.14.2-py2.py3-none-any.whl"
+)
```
alanag13, line 7: just `black` again

tests/test_detectionlists_actions.py
```diff
@@ -1,12 +1,18 @@
 from pytest import fixture
 
-from tests.conftest import assert_success, create_fake_connector, create_mock_response, assert_successful_single_data, assert_successful_message, assert_succesful_summary
+from tests.conftest import (
```
alanag13, line 3: all changes to the tests are just `black`

antazoey:
These build scripts might go away some day or become out of date

CONTRIBUTING.md
```diff
@@ -14,38 +14,54 @@ Also confirm that you can access the Phantom web app at `https://{phantom_VM_IP_
 
 NOTE: If you are consistently getting timeouts connecting to the Phantom VM, try disconnecting from the VPN and re-attempting.
 
+### Style linter
+
+When you open a PR a series of style checks will run. See the [pre-commit-config.yaml](.pre-commit-config.yaml) file to see a list of the projects involved in this automation. If your code does not pass the style checks, the PR will not be allowed to merge. Many of the style rules can be corrected automatically by running a simple command once you are satisfied with your change:
```
antazoey: comma after `when you open a PR`

CONTRIBUTING.md
```diff
@@ -14,38 +14,54 @@ Also confirm that you can access the Phantom web app at `https://{phantom_VM_IP_
...
+
+```bash
+make style
+```
+
+This will output a diff of the files that were changed. Once these have been corrected and re-pushed, the PR checks should pass.
+
+You can optionally also choose to have these checks / automatic adjustments
```
antazoey: Probably don't need the both words `optionally` and `also` at the beginning of this sentence

CONTRIBUTING.md
```diff
@@ -93,15 +109,28 @@ To leave the virtual environment, simply use:
...
 pip install -e .[dev]
-pytest
+make
+```
+
+By default, `make` will run all unit tests, Phantom configuration / lint validators, style checks, and create a local tarball of the app.
+
+Each of these commands can also be run individually if desired:
+
+```bash
```
antazoey, line 126: I am a little sad you removed the `ssh` and the `open-web` commands, but I can make custom aliases for those if you don't think they belong. I was using them though.

antazoey:
Could alternatively make a command that pulls the scripts off the server?

Edit: I guess that wouldn't work for CI/CD

CONTRIBUTING.md
alanag13: I didn't mean to leave them out! I can add them back, sorry about this.

alanag13:
@unparalleled-js 

> These build scripts might go away some day or become out of date

If we ever have to build against a different, newer version of Phantom, sure. Until then they won't just update under our noses. If we want to be able to leverage them in CI/CD, they have to be checked in because the pipeline won't be able to pull them from a server.

CONTRIBUTING.md
alanag13: back in now!

Makefile
```diff
@@ -28,3 +28,9 @@ deploy-bypass:: tar
 	sshpass -p "${PHANTOM_VM_PASSWORD}" scp ../phcode42v2.tgz "phantom@${PHANTOM_VM_IP_ADDR}":/home/phantom
 	sshpass -p "${PHANTOM_VM_PASSWORD}" ssh phantom@${PHANTOM_VM_IP_ADDR} \
 		"rm -rf phcode42v2 && tar -xvf phcode42v2.tgz && cd phcode42v2 && phenv python /opt/phantom/bin/compile_app.pyc -i"
+
+ssh::
+	sshpass -p "${PHANTOM_VM_PASSWORD}" ssh phantom@${PHANTOM_VM_IP_ADDR}
+
+open-web::
+	open https://${PHANTOM_VM_IP_ADDR}:9999
```
unparalleled-js: How come the linter does not care that there is a missing newline here?

antazoey (approved):
I am considering the .pyc files to be temporary.
It seems like .py files for these tasks are in the horizon, based on what I have seen in slack

alanag13:
> It seems like .py files for these tasks are in the horizon, based on what I have seen in slack

Good, I would definitely much rather have something that feels a little more official / supported than this -- but who knows what "on the horizon" means :-(

peterbriggs42 (approved):

PR #11, opened 2021-05-26
Feature/alerts

peterbriggs42:


tests/test_alert_actions.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def test_handle_action_when_search_alerts_and_no_date_range_provided_defaults_to_last_30_days(
+        self, mock_py42_client
+    ):
+        param = {"username": "barney.frankenberry@chocula.com"}
+        connector = _create_search_alerts_connector(mock_py42_client)
+        connector.handle_action(param)
+        actual_query = mock_py42_client.alerts.search.call_args[0][0]
+        # This is not a great assertion...
+        assert len(actual_query._filter_group_list) == 2
```
peterbriggs42: Wasn't able to figure out a clean way to assert that a default "last 30 days" `DateObserved` filter is applied when no date range params are provided. Tried one way involving accessing internal members on the `AlertsQuery` object but it got ugly fast...

Any ideas on how/whether to improve?

code42.json
```diff
@@ -629,6 +629,331 @@
...
+                    "default": "",
+                    "order": 0
+                }
+            },
+            "output": [
+                {
+                    "data_path": "action_result.parameter.alert_id",
+                    "data_type": "string",
+                    "contains": [
+                        "alert"
```
antazoey: I don't see this in the list of built-in contains types. Are we able to use custom ones out of the box like this?

code42.json
```diff
@@ -629,6 +629,331 @@
...
+                {
+                    "data_path": "action_result.status",
+                    "data_type": "string",
+                    "column_name": "Status",
+                    "column_order": 1
+                },
+                {
+                    "data_path": "action_result.data.*.username",
+                    "data_type": "string",
+                    "column_name": "username",
```
antazoey, line 1103: Could you add the `email` contains there so this plugs in to the other actions easily?

code42.json
```diff
@@ -629,6 +629,331 @@
...
+                },
+                {
+                    "data_path": "action_result.data.*.name",
+                    "data_type": "string",
+                    "column_name": "Name",
+                    "column_order": 8
+                },
+                {
+                    "data_path": "action_result.summary.username",
+                    "data_type": "string",
```
antazoey, line 1147: Could also add the email contains here in case someone for some reason want to use the summary value

code42.json
```diff
@@ -629,6 +629,331 @@
...
+                "type": "table"
+            },
+            "versions": "EQ(*)"
+        },
+        {
+            "action": "set alert state",
+            "identifier": "set_alert_state",
+            "description": "Set the state of the alert",
+            "verbose": "",
+            "type": "investigate",
```
antazoey: Should this type be different, like `contain` ?

code42.json
```diff
@@ -629,6 +629,331 @@
...
+            "read_only": false,
+            "parameters": {
+                "alert_id": {
+                    "description": "Alert ID",
+                    "data_type": "string",
+                    "required": true,
+                    "primary": true,
```
antazoey, line 1319: If you are going to use the `alert` contains like you did above for the `alert ID`, it should be used here as well (and everywhere `alert id` is defined

code42_connector.py
```diff
@@ -154,11 +166,90 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+        return action_result.set_status(phantom.APP_SUCCESS)
+
+    @action_handler_for("search_alerts")
+    def _handle_search_alerts(self, param, action_result):
+        username = param.get("username")
+        start_date = param.get("start_date")
+        end_date = param.get("end_date")
+        alert_state = param.get("alert_state")
+
+        if username is None and start_date is None and end_date is None and alert_state is None:
```
antazoey: Could be `if not username and not start_date and not end_date and not alert_state`, looks a little simpler and I don't think any of these params should work for falsy values

code42_connector.py
```diff
@@ -154,11 +166,90 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+        status_message = f"State of alert {alert_id} was updated to {alert_state}"
+        action_result.update_summary({"alert_id": alert_id})
+        return action_result.set_status(phantom.APP_SUCCESS, status_message)
+
     def finalize(self):
         # Save the state, this data is saved across actions and app upgrades
         self.save_state(self._state)
         return phantom.APP_SUCCESS
 
+    def _validate_date_range(self, start_date, end_date):
```
unparalleled-js, line 230: The name of this method sounds like it is validating the date range consists of enough days or something, perhaps it should be named `_validate_date_args()`

code42_connector.py
```diff
@@ -154,11 +166,90 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+        return (not start_date and not end_date) or (
```
unparalleled-js, line 231: It took me quite a while for some reason to understand what was going on here, lots of really simpler logic gates....

Maybe just a comment like `# Requires both args if present` or something

code42_connector.py
```diff
@@ -154,11 +166,90 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+
+        if username is None and start_date is None and end_date is None and alert_state is None:
+            return action_result.set_status(
+                phantom.APP_ERROR, "Must supply a search term."
+            )
+
+        if not self._validate_date_range(start_date, end_date):
+            return action_result.set_status(
+                phantom.APP_ERROR,
+                "Start Date and End Date are both required to search by " "date range.",
```
antazoey: This string is oddly split in two

code42.json
peterbriggs42: Oops, that was a mistake. Removed

code42_connector.py
```diff
@@ -154,11 +166,90 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+
+        if username is None and start_date is None and end_date is None and alert_state is None:
+            return action_result.set_status(
+                phantom.APP_ERROR, "Must supply a search term."
+            )
+
+        if not self._validate_date_range(start_date, end_date):
+            return action_result.set_status(
+                phantom.APP_ERROR,
+                "Start Date and End Date are both required to search by date range.",
```
antazoey: hmm... In other apps, if the user only provides a begin, we use the default for the end, and visa-versa

tests/test_alert_actions.py
alanag13: It's been a while, but I think you could convert it to a dict and read it that way via `dict(actual_query)`.  If I remember right, that object has a special overload that makes it get converted to its JSON representation when you do that.

code42_connector.py
alanag13: Yeah, typically just the start date means "from that time until now" and just the end date means "everything before this time".

tests/test_alert_actions.py
antazoey: As far as the assertion goes, I typically create an expected date (one 30 days back in this case) and assert that the actual date is equivalent (like subtract them, take absolute value, assert `< 0.01` or something similar).

As far as accessing the value, you can call `str()` on the query - that converts it to JSON and you can parse a `dict` out of it using `json.loads(json_str)` and then access the value that way

code42_connector.py
```diff
@@ -154,11 +166,90 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+    @action_handler_for("search_alerts")
+    def _handle_search_alerts(self, param, action_result):
+        username = param.get("username")
+        start_date = param.get("start_date")
+        end_date = param.get("end_date")
+        alert_state = param.get("alert_state")
+
+        if username is None and start_date is None and end_date is None and alert_state is None:
+            return action_result.set_status(
+                phantom.APP_ERROR, "Must supply a search term."
```
alanag13: To make logs easier to parse in a land where users will probably have multiple integrations, I've been trying to include `Code42` and the action names in the messages. Something like:

`Code42: Must supply a search term when calling action 'search_alerts'.`

code42_connector.py
```diff
@@ -154,11 +166,90 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+        )
+
+    def _build_alerts_query(self, username, start_date, end_date, alert_state):
+        filters = []
+        if username is not None:
+            filters.append(Actor.eq(username))
+        if start_date is not None and end_date is not None:
+            # This will raise a ValueError if dates not in Y-m-d H:M:S format
+            filters.append(
+                DateObserved.in_range(f"{start_date} 00:00:00", f"{end_date} 00:00:00")
```
alanag13: does phantom only let you input calendar dates (no timestamps)? Trying to understand the `00:00:00` here.

tests/test_alert_actions.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def test_handle_action_when_set_alert_state_calls_update_state(
```
alanag13, line 221: for tests like this, can you include an additional `assert_success`? This is a pattern I tried to follow with all the other tests. It's valuable to know that in addition to the function being called with the expected params, its return status was set correctly. There's a couple tests in here that could benefit from this.

Basically, every test should use at least one of the helper assertion methods that is in `conftest.py`.

tests/test_alert_actions.py
```diff
@@ -0,0 +1,242 @@
...
+            "username": "gilbert.lequiche@finance.biz",
+            "start_date": "2031-01-01",
+            "end_date": "2031-06-01",
+            "alert_state": "OPEN",
+        }
+        connector = _create_search_alerts_connector(mock_py42_client)
+        connector.handle_action(param)
+        mock_py42_client.alerts.search.assert_called_once()
+
+    def test_handle_action_when_search_alerts_and_all_params_missing_sets_error_message(
```
alanag13, line 158: These tests are all titled that they are testing `sets_error_message`, but they all just check the _status_ is an error. Might be a good time to add an `assert_failure_message` helper?

code42_connector.py
peterbriggs42: Good call. Wouldn't have thought of that

code42_connector.py
timabrmsn: could also do:

```
if not all(username, start_date, end_date, alert_state):
```

code42_connector.py
peterbriggs42: No Phantom doesn't have a dedicated Date or DateTime input so it's just a string. The requirement just said to allow user to supply a date range so I figured that exposing time-level filtering was more granular than the usecase required. Now that I think about it, security analysts searching for alerts probably *do* require that kind of granularity when building out queries.

If we allow the `start_date` and `end_date` parameters to be entered with the timestamp included, should we require that formatting or also allow them to omit the time and only supply date?

code42_connector.py
peterbriggs42: Alright cool, I can change the behavior to align better with the other apps.

code42_connector.py
```diff
@@ -154,11 +166,90 @@ def _handle_list_high_risk_employees(self, param, action_result):
...
+
     def finalize(self):
         # Save the state, this data is saved across actions and app upgrades
         self.save_state(self._state)
         return phantom.APP_SUCCESS
 
+    def _validate_date_range(self, start_date, end_date):
+        return (not start_date and not end_date) or (
+            start_date is not None and end_date is not None
+        )
```
timabrmsn: Can we clarify this function? I'm not sure what makes a date range "valid" from this logic.

code42_connector.py
timabrmsn: it looks like the phantom env has the [python-dateutil](https://pypi.org/project/python-dateutil/) package installed, we could just call `dateutil.parser.parse()` on the strings and if it fails say it's not a recognized date format. Then we don't have to do any manual conversions/parsing.

alanag13: That definitely seems more flexible/safe. The comment that we have in the code right now is also misleading (I'm pretty sure something bad would happen if we included `H:M:S` in the input right now since `00:00:00` would still be concat'd on).

code42_connector.py
```diff
@@ -225,11 +229,91 @@ def _handle_remove_high_risk_tag(self, param, action_result):
...
+        action_result.update_summary({"total_count": response["totalCount"]})
+        return action_result.set_status(phantom.APP_SUCCESS)
+
+    @action_handler_for("set_alert_state")
+    def _handle_set_alert_state(self, param, action_result):
+        alert_id = param["alert_id"]
+        alert_state = param["alert_state"]
+        note = param.get("note")
+        response = self._client.alerts.update_state(alert_state, [alert_id], note=note)
+        action_result.add_data(response.data)
```
peterbriggs42, line 270: Getting this output from compile_app.pyc for the `set_alert` JSON definition, it seems to be complaining that I'm not exposing any properties on `action_result.data` in the output array:
```    
set alert state
      Following required data paths not in output list
        action_result.data
```
But the `py42.alerts.update_state` method doesn't actually return any response data, and there's nothing that I've calculated in the action itself that isn't already present in the parameters. It feels bad to programmatically add something useless to `action_result.add_data`. But maybe it's worth it to suppress the compiler warning?

Thoughts?

code42_connector.py
peterbriggs42: Changed how date queries work here and removed this function

code42_connector.py
peterbriggs42: That's a great idea Tim. I had found that package when I was first researching this problem but was reluctant to add a new dependency for such a little thing. Much cleaner and more flexible now that we're using the `dateutil` parser.

unparalleled-js: I think the API technically allows any format, right?
We shouldn't make it harder to input dates - what if the date is coming from some other app and it is a longer format? It'd be nice if we could just pass it through to this method without the playbook author needing to worry about it

code42_connector.py
antazoey: What does the generated action JSON look like? Does it have any sort of placeholder for `.data`?

code42_connector.py
```diff
@@ -225,11 +230,76 @@ def _handle_remove_high_risk_tag(self, param, action_result):
...
+        filters.append(self._build_date_range_filter(start_date, end_date))
+        query = AlertQuery.all(*filters)
+        return query
+
+    def _build_date_range_filter(self, start_date, end_date):
+        if start_date and not end_date:
+            return DateObserved.on_or_after(dateutil.parser.parse(start_date))
+        elif end_date and not start_date:
+            return DateObserved.on_or_before(dateutil.parser.parse(end_date))
+        elif end_date is not None and start_date is not None:
```
antazoey: `not end_date and not start_date` ?

code42_connector.py
peterbriggs42: Spun up a dummy action and found that the auto-generated action JSON also has this issue
```    search
      Action name "search" is invalid. Action names should be 2 or 3 words long separated by a space
      Data Path "action_result.parameter.username" has invalid contains, A contains list was provided, but it was empty.
      Data Path "action_result.parameter.state" has invalid contains, A contains list was provided, but it was empty.
      Following required data paths not in output list
        action_result.data
        action_result.summary
```

antazoey (approved):

PR #12, opened 2021-05-28
Force use of pyenv python bins

antazoey:
* Force use pyenv specific binaries.
I am having issues with pyenv and the python binaries that are in my /usr/bin, this makes it always use the right binaries.
Also, I added a note in the contributing about an alternative way python 3.6.13 to install in case of troubles

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,12 @@ pyenv activate phantom
 pip install --upgrade pip
 ```
 
+Note: if the command `pyenv virtualenv 3.6.13 phantom` fails, try the following command instead:
+
+```bash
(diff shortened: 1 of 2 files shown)


antazoey:
CI/CD does not use pyenv.
Thinking...

antazoey:
Trying to resolve:

```
pyenv activate phenv
which python
>> /usr/bin/python
```

antazoey:
I have tried commenting out my entire bash profile

antazoey:
I guess the other alternative is do 
`alias python="$(pyenv which python)"`
for each binary myself locally...

antazoey:
I give up lol

PR #16, opened 2021-06-01
Implement legal hold actions

peterbriggs42:


code42.json
```diff
@@ -1746,6 +1746,171 @@
                 "type": "table"
             },
             "versions": "EQ(*)"
+        },
+        {
+            "action": "add legalhold user",
```
antazoey: Can we call these actions `add to matter` and `remove from matter` instead? Thoughts?

code42_connector.py
```diff
@@ -367,6 +390,27 @@ def _get_user(self, username):
     def _get_user_id(self, username):
         return self._get_user(username)["userUid"]
 
+    # Following two helper functions are copy+pasted from cmds/legal_hold.py in `code42cli`
+    def _get_legal_hold_membership_id_for_user_and_matter(self, user_id, matter_id):
+        memberships = self._get_legal_hold_memberships_for_matter(matter_id)
+        for member in memberships:
+            if member["user"]["userUid"] == user_id:
+                return member["legalHoldMembershipUid"]
+        raise ValueError(
```
antazoey: Is `ValueError` the correct error here? `ValueError` makes it seem like the method was given a username instead of an id, or some other similar scenario.

Since this is just behind-the-scenes logic, it might make more sense to return `None` from this method. Then, in the action handlers, check if the id is `None` and fail accordingly (rather than `try` / `except`).

tests/test_legalhold_actions.py
```diff
@@ -0,0 +1,170 @@
...
+    connector = create_fake_connector("add_legalhold_user")
+    return attach_client(connector, client)
+
+
+def _create_remove_legalhold_user_connector(client):
+    connector = create_fake_connector("remove_legalhold_user")
+    return attach_client(connector, client)
+
+
+class TestCode42AlertsConnector(object):
```
antazoey: This class should have `LegalHold` in the name rather than `Alerts`

tests/test_legalhold_actions.py
```diff
@@ -0,0 +1,170 @@
...
+            "legalHoldMembershipUid": "1009118679333039731",
+            "active": True,
+            "creationDate": "2021-05-28T19:13:24.312Z",
+            "legalHold": {
+                "legalHoldUid": "1002360211054033697",
+                "name": "test-matter-752367781",
+            },
+            "user": {
+                "userUid": "990572034162882387",
+                "username": "peter.briggs+partner@code42.com",
```
antazoey: Can we sanitize real user emails?

tests/test_legalhold_actions.py
```diff
@@ -0,0 +1,170 @@
...
+        assert_successful_message(
+            connector,
+            f"test@example.com was added to legal hold matter {_TEST_MATTER_ID}.",
+        )
+
+    def test_handle_action_when_remove_legal_hold_user_calls_remove_with_expected_args(
+        self, mock_py42_with_legal_hold_memberships
+    ):
+        param = {"username": "test@example.com", "matter_id": _TEST_MATTER_ID}
+        # mock_py42_with_legal_hold_memberships.legalhold.get_all_matter_custodians
```
antazoey: Commented out code?

code42.json
alanag13: maybe `add user to legal hold matter` and `remove user from legal hold matter`?

code42.json
```diff
@@ -1746,6 +1746,171 @@
...
+            "identifier": "add_legalhold_user",
+            "description": "Add a user (custodian) to a legal hold matter",
+            "verbose": "",
+            "type": "contain",
```
alanag13, line 1756: It probably makes sense to make use of the `undo` functionality here like we do with the detection lists.

code42.json
peterbriggs42: @alanag13 there's a 3-word limit on action names :(
@unparalleled-js I was following the pattern of `add highrisk employee` and `add departing employee`. I feel like `add to matter` is not quite as informative about what type of entity is being added to the matter. And `add matter user` just reads kind of clunky to me, though if you think that's preferable I'd be ok with it.

tests/test_legalhold_actions.py
peterbriggs42: Oops, thought I cleaned that one up. Removed.

code42.json
antazoey: I agree the word limit causes this to be less than ideal / uninformative... How about this:

`add_legalhold_custodian`. I think that sums it up nicely and implies "matter".

Thoughts?

peterbriggs42: Ahh yep I like that. Will change

antazoey (approved):

PR #19, opened 2021-06-04
Feature/file event actions

alanag13:
adds stream-by-hash and file event searching actions.

The stream-by-hash functions will have their files land on the "Files" tab in an event.
The output for both the advanced and "normal" file event query is the same, and the table structure I chose is based on what the default columns are in the FFS UI.

The code42.json and unit tests file are hidden by default due to large diff so make sure to look at those.


code42_connector.py
```diff
@@ -40,6 +64,17 @@ def _convert_to_obj_list(scalar_list, sub_object_key="item"):
     return [{sub_object_key: item} for item in scalar_list]
 
 
+def add_eq_filter(filters, value, filter_class):
+    return filters.append(filter_class.eq(value))
+
+
+def is_default_dict(_dict):
```
alanag13, line 71: this gets shared between alerts and file events - you can use it to check to see if all the params were left blank

code42_connector.py
```diff
@@ -370,22 +458,85 @@ def _build_alerts_query(self, username, start_date, end_date, alert_state):
...
+        filters.append(
+            self._build_date_range_filter(EventTimestamp, start_date, end_date)
+        )
+        query = FileEventQuery.all(*filters)
+        return query
+
+    def _build_date_range_filter(self, date_filter_cls, start_date, end_date):
         if start_date and not end_date:
-            return DateObserved.on_or_after(dateutil.parser.parse(start_date))
+            return date_filter_cls.on_or_after(dateutil.parser.parse(start_date))
```
alanag13, line 537: modified this func slightly to make it so that it could also be used with `EventTimestamp`

stubs/phantom/utils.py
```diff
@@ -0,0 +1,13 @@
+import re
+
+
+def is_md5(input_str):
```
alanag13, line 4: these functions are build in to phantom so they're stubbed here

stubs/phantom/vault.py
```diff
@@ -0,0 +1,6 @@
+class Vault:
+    @classmethod
+    def create_attachment(
```
alanag13, line 3: this is the method that adds raw bytes as an attachment under the `Files` tab

code42.json
```diff
@@ -1924,6 +1924,964 @@
                 "type": "table"
             },
             "versions": "EQ(*)"
+        },
+        {
+            "action": "hunt file",
```
alanag13, line 1929: these names (`hunt file` and `run query`) are based off the name from the old app. I remember Phantom specifically saying that they try to name certain types of actions with the same action name because this makes them get grouped together in the UI (i.e. you could pick "hunt file" and then choose which app you want to hunt the file in).

code42_connector.py
```diff
@@ -359,6 +396,57 @@ def _handle_remove_legal_hold_custodian(self, param, action_result):
         status_message = f"{username} was removed from legal hold matter {matter_id}."
         return action_result.set_status(phantom.APP_SUCCESS, status_message)
 
+    """ FILE EVENT ACTIONS """
+
+    @action_handler_for("hunt_file")
+    def _handle_hunt_file(self, param, action_result):
```
alanag13, line 405: There's only one stream file action and it just picks the right method based on detecting the type of hash

code42.json
```diff
@@ -1924,6 +1924,964 @@
...
+            "type": "investigate",
+            "read_only": true,
+            "parameters": {
+                "hash": {
+                    "description": "The md5 or sha256 hash of the file to download",
+                    "data_type": "string",
+                    "required": true,
+                    "primary": true,
+                    "contains": [
+                        "md5",
```
alanag13, line 1942: I didn't use `hash` because some decompiled stuff that I saw suggested that that would make this allowed to be attempted on other types of hashes that we don't support like `SHA1`.

tests/test_file_event_actions.py
```diff
@@ -0,0 +1,615 @@
...
+    def test_handle_action_when_hunt_file_given_unsupported_hash_sets_error_message(
+        self, mocker, mock_py42_client
+    ):
+        param = {"hash": "not-a-valid-hash"}
+        connector = _create_hunt_file_connector(mocker, mock_py42_client)
+        connector.handle_action(param)
+        message = "Code42: Failed execution of action hunt_file: Unsupported hash format. Hash must be either md5 or sha256"
+        assert_fail_message(connector, message)
+
+    def test_handle_action_when_run_query_calls_search_file_events_with_expected_query(
```
alanag13, line 463: checks to make sure each value that is passed in actually shows up in the query.

code42.json
```diff
@@ -1924,6 +1924,964 @@
...
+                        "sha256"
+                    ],
+                    "default": "",
+                    "order": 0
+                },
+                "file_name": {
+                    "description": "The name to give to the file after it is downloaded. If not supplied, the hash will be used.",
```
antazoey, line 1949: But hash is a required argument?  Does that mean if I supply a file name, it ignores the hash? If that is the case, perhaps hash should not be required

code42.json
```diff
@@ -1924,6 +1924,964 @@
...
+                    "description": "File path",
+                    "data_type": "string",
+                    "order": 4,
+                    "primary": true,
+                    "contains": [
+                        "file path"
+                    ]
+                },
+                "file_category": {
+                    "description": "File category",
```
antazoey: Could make these description slightly better by becoming inspired from the py42 docs: https://py42docs.code42.com/en/stable/methoddocs/filleeventqueries.html#py42.sdk.queries.fileevents.filters.file_filter.FileCategory

^ In this case, something like `Filter events by category of the file observed`.

code42.json
```diff
@@ -1924,6 +1924,964 @@
...
+                    "data_type": "string",
+                    "column_name": "SHA256 hash",
+                    "column_order": 7,
+                    "contains": [
+                        "sha256"
+                    ]
+                },
+                {
+                    "data_path": "action_result.data.*.url",
+                    "data_type": "string"
```
antazoey: Could this one also have the `url` contains type, or no?

code42.json
```diff
@@ -1924,6 +1924,964 @@
...
+                    "data_path": "action_result.data.*.fileType",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "action_result.data.*.deviceUid",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "action_result.data.*.emailFrom",
+                    "data_type": "string"
```
antazoey: Email contains type works here I think

code42.json
```diff
@@ -1924,6 +1924,964 @@
...
+                {
+                    "data_path": "action_result.data.*.osHostName",
+                    "data_type": "string",
+                    "contains": [
+                        "host name"
+                    ]
+                },
+                {
+                    "data_path": "action_result.data.*.emailSender",
+                    "data_type": "string"
```
antazoey: email contains type could work here

code42.json
```diff
@@ -1924,6 +1924,964 @@
...
+        {
+            "action": "run advanced query",
+            "identifier": "run_advanced_query",
+            "description": "Run an advanced query using JSON",
+            "verbose": "",
+            "type": "investigate",
+            "read_only": true,
+            "parameters": {
+                "json_query": {
+                    "description": "The md5 or sha256 hash of the file to download",
```
antazoey: I think this description is wrong

code42.json
```diff
@@ -1924,6 +1924,964 @@
...
+                {
+                    "data_path": "action_result.data.*.fileType",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "action_result.data.*.deviceUid",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "action_result.data.*.emailFrom",
```
antazoey, line 2706: Shared feedback for both action outputs jsons.. Too bad we can't reuse models.

code42.json
alanag13: No, the file name isn't used in search at all, it only controls what the file should be named after it's downloaded. The hash is required and is the only thing used to look up the file. What this is trying to say is that if you don't supply a name for the file, the hash will be the name given to it.

code42.json
alanag13: I pulled this pattern from the existing app but have no problem with making them a little more detailed

code42.json
alanag13: lololol quite

code42_connector.py
antazoey: `has_no_values` might be a better name.

code42_connector.py
```diff
@@ -40,6 +64,17 @@ def _convert_to_obj_list(scalar_list, sub_object_key="item"):
     return [{sub_object_key: item} for item in scalar_list]
 
 
+def add_eq_filter(filters, value, filter_class):
+    return filters.append(filter_class.eq(value))
+
+
+def is_default_dict(_dict):
+    for key in _dict:
+        if _dict.get(key):
```
antazoey: Would `{"foo": False}` be a false positive here?
(pun intended)

code42_connector.py
```diff
@@ -359,6 +396,57 @@ def _handle_remove_legal_hold_custodian(self, param, action_result):
         status_message = f"{username} was removed from legal hold matter {matter_id}."
         return action_result.set_status(phantom.APP_SUCCESS, status_message)
 
+    """ FILE EVENT ACTIONS """
+
+    @action_handler_for("hunt_file")
+    def _handle_hunt_file(self, param, action_result):
+        file_hash = param["hash"]
+        file_name = param.get("file_name")
```
antazoey, line 407: This whole part could be

```
file_name = param.get("file_name", file_hash)
param["file_name"] = file_name
```

however, that might be more confusing

antazoey: oh, I guess the `update_param` doesn't fit there ... nevermind!

code42_connector.py
alanag13: `False` would be the default for any checkbox item. Currently the only one of these is `untrusted_only`. If that's unchecked, and there is nothing else set, that means "get all the events", which is exactly what we want to avoid -- so in the current state of the app there's nothing wrong with this behavior.

antazoey: You can change the default options I am pretty sure. I see what you mean though, regarding trusted_activity being the only one and wanting to avoid getting all the results...

I think it is probably fine for now, but it would be better to be able instead check that each individual parameter is not its own default. I am not sure if the input schema contains that info or not or if there is any other way to know.

code42_connector.py
antazoey: Nevermind, now I see what it is called this

code42.json
antazoey: Yes, I remember this from XSOAR now, got it

tests/test_file_event_actions.py
```diff
@@ -0,0 +1,626 @@
...
+    def test_handle_action_when_run_query_and_date_format_invalid_sets_error_status(
+        self, mocker, mock_py42_client
+    ):
+        param = {
+            "start_date": "2031-31-12",
+            "end_date": "2032-31-01",
+        }
+        connector = _create_run_query_connector(mocker, mock_py42_client)
+        connector.handle_action(param)
+        assert_fail(connector)
```
antazoey, line 542: I am curious what the failure message looks like in this case

tests/test_file_event_actions.py
```diff
@@ -0,0 +1,626 @@
...
+        )
+        assert_successful_message(connector, message)
+
+    def test_handle_action_when_hunt_file_given_unsupported_hash_sets_error_message(
+        self, mocker, mock_py42_client
+    ):
+        param = {"hash": "not-a-valid-hash"}
+        connector = _create_hunt_file_connector(mocker, mock_py42_client)
+        connector.handle_action(param)
+        message = "Code42: Failed execution of action hunt_file: Unsupported hash format. Hash must be either md5 or sha256"
```
antazoey: I think the equivalent of this test is missing for the query action

antazoey: Done

tests/test_file_event_actions.py
antazoey: I added a comment.


I chose a comment rather than `assert_failure_message()` since the error message in this case comes from date util or something.. it looks decent though and explains the problem well enough, which was what I wanted to make sure of.

code42.json
antazoey: Done!

code42.json
antazoey: Done

code42.json
antazoey: Done

code42.json
antazoey: Done

code42.json
peterbriggs42: It's still wrong I think

timabrmsn:
Would it make sense to have the `file event` (ID) be a "contains" parameter to be able to add to a case from search results?

code42_connector.py
```diff
@@ -359,6 +396,57 @@ def _handle_remove_legal_hold_custodian(self, param, action_result):
...
+        file_content = self._get_file_content(file_hash)
+        container_id = self.get_container_id()
+        Vault.create_attachment(file_content, container_id, file_name=file_name)
+        status_message = f"{file_name} was successfully downloaded and attached to container {container_id}"
+        return action_result.set_status(phantom.APP_SUCCESS, status_message)
+
+    @action_handler_for("run_query")
+    def _handle_run_query(self, param, action_result):
+        if is_default_dict(param):
+            return action_result.set_status(
```
peterbriggs42, line 426: If you omit all the search terms, we will still run the search because the `untrusted_only` flag has a default value of `False`. So this failure block will never be hit when running the action.

peterbriggs42 (commented):
It looks like the requirement for `run query` action to return a max of 10k results was not implemented.

tests/test_file_event_actions.py
```diff
@@ -0,0 +1,655 @@
...
+
+        check_filter(query_json, "ON_OR_AFTER", "2031-01-01T00:00:00.000Z")
+        check_filter(query_json, "ON_OR_BEFORE", "2031-06-01T00:00:00.000Z")
+
+        assert_success(connector)
+
+    def test_handle_action_when_run_query_and_all_params_missing_or_default_sets_error_message(
+        self, mocker, mock_py42_client
+    ):
+        param = {"untrusted_only": False}
```
antazoey: @peterbriggs42

code42_connector.py
antazoey: See https://github.com/phantomcyber/phcode42v2/pull/19/files#r645792039

Also, I just adjusted the test to have the default value and the failure block is still hit. (https://github.com/phantomcyber/phcode42v2/pull/19/files#r646798948)

`is_default_dict` checks if the values are their defaults, basically, via falsy-ness.

Unless I am misunderstanding something?

antazoey:
> It looks like the requirement for `run query` action to return a max of 10k results was not implemented.

It does this automatically in the API https://py42docs.code42.com/en/stable/methoddocs/securitydata.html?highlight=search_file_events#py42.clients.securitydata.SecurityDataClient.search_file_events

antazoey:
> Would it make sense to have the `file event` (ID) be a "contains" parameter to be able to add to a case from search results?

I don't fully understand custom contains type (I don't see `file event` as a type listed in `/rest/cef_metadata` or the UI).
The answer is yes though! I just don't know when to do this.

You _are_ able to pass params around in Playbooks without sharing the contains types, you just have to click an extra menu button to see all the possible ones.

timabrmsn:
> > Would it make sense to have the `file event` (ID) be a "contains" parameter to be able to add to a case from search results?
> 
> I don't fully understand custom contains type (I don't see `file event` as a type listed in `/rest/cef_metadata` or the UI).
> The answer is yes though! I just don't know when to do this.
> 
> You _are_ able to pass params around in Playbooks without sharing the contains types, you just have to click an extra menu button to see all the possible ones.

It seems creating your own contains types is encouraged, but best to use a pre-existing one if possible as it'll work with other apps that use the same ones.

From https://docs.splunk.com/Documentation/Phantom/4.10.3/DevelopApps/Contains :

>This is a powerful feature that the platform provides, as it allows the user to chain the output of one action as input to another. As an app author, check that your data type isn't already covered by an existing contains that other apps use before creating a new one for their app.

antazoey (approved):
When you all approve, let's merge!

code42_connector.py
peterbriggs42: I think that leaving the `untrusted_only` checkbox unchecked will supply a value of `'false'` to the `params` dict, rather than the Python bool value of `False`. When I run the action without supplying any parameter values in the UI I get 57k results returned back to me.

Edit: I actually only get 10k results returned; forgot that `totalCount` reflects the total number of results for the query on the backend, not the number of results returned in the response.

code42_connector.py
```diff
@@ -398,6 +549,24 @@ def _get_user(self, username):
...
 
+    def _get_file_content(self, file_hash):
+        if utils.is_md5(file_hash):
+            response = self._client.securitydata.stream_file_by_md5(file_hash)
+        elif utils.is_sha256(file_hash):
+            response = self._client.securitydata.stream_file_by_sha256(file_hash)
+        else:
+            raise Code42UnsupportedHashError()
+
+        chunks = [chunk for chunk in response.iter_content(chunk_size=128) if chunk]
```
timabrmsn, line 567: Is chunking even necessary here since we're not doing anything in between the chunks? We should just be able to return `r.content` straight away, no?

antazoey: I think we have to do this because of this line: https://github.com/code42/py42/blob/master/src/py42/services/storage/preservationdata.py#L50

When `stream=False`, we are able to use the `content` property, otherwise the content is not returned in full.. At least according to https://py42docs.code42.com/en/stable/methoddocs/response.html#py42.response.Py42Response.iter_content

antazoey: Although, now that you bring this up, we could potentially have a very large file in memory at this point. I am not sure if that is avoidable or not, I have no idea how this part of Phantom works yet

code42_connector.py
```diff
@@ -361,6 +399,61 @@ def _handle_remove_legal_hold_custodian(self, param, action_result):
...
+        container_id = self.get_container_id()
+        Vault.create_attachment(file_content, container_id, file_name=file_name)
+        status_message = f"{file_name} was successfully downloaded and attached to container {container_id}"
+        return action_result.set_status(phantom.APP_SUCCESS, status_message)
+
+    @action_handler_for("run_query")
+    def _handle_run_query(self, param, action_result):
+        # Boolean action parameters are passed as lowercase string representations, fix that here.
+        if "untrusted_only" in param:
+            param["untrusted_only"] = param["untrusted_only"] == "true"
```
antazoey: Can we change line 423 to be

```python
param["untrusted_only"] = param["untrusted_only"].lower() in [True, "true"]
```

peterbriggs42 (approved):

PR #20, opened 2021-06-07
Check whether matter ID is accessible before attempting to add/remove custodian

peterbriggs42:
From Nikunj's testing of legal hold actions:
```
When I try to add legal hold custodian with invalid Matter Id it gives me below error :
Code42: Failed execution of action add_legalhold_custodian: Failure in HTTP call 403 Client Error: Forbidden for url: https://console.us.code42.com/api/LegalHoldMembership. Response content: [{"name":"SYSTEM","description":"Not authorized for this action"}]
```

This PR addresses the issue by checking whether the matter is accessible given current Py42 profile before adding or removing custodians on it.

### DIFF
--- a/code42_connector.py
+++ b/code42_connector.py
@@ -419,6 +419,7 @@ class Code42Connector(BaseConnector):
         ]
         return memberships
 
+    # Fails when a legal hold matter is inaccessible from the user's account or the matter ID is not valid
     def _check_matter_is_accessible(self, matter_id):
         return self._client.legalhold.get_matter_by_uid(matter_id)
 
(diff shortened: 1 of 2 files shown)


code42_connector.py
```diff
@@ -417,6 +419,9 @@ def _get_legal_hold_memberships_for_matter(self, matter_id):
         ]
         return memberships
 
+    def _check_matter_is_accessible(self, matter_id):
+        return self._client.legalhold.get_matter_by_uid(matter_id)
```
antazoey, line 424: Might be good to include comment here saying 

```python
# Fails when a legal hold matter is inaccessible from the user's account or the matter ID is not valid
```

antazoey (approved):

PR #21, opened 2021-06-08
Feature/cases actions

timabrmsn:


code42.json
```diff
@@ -1924,6 +1924,463 @@
...
+                },
+                "description": {
+                    "description": "Description",
+                    "data_type": "string",
+                    "required": false,
+                    "primary": false,
+                    "default": "",
+                    "order": 2
+                },
+                "assignee": {
```
antazoey, line 1963: Should we add contains types for `assignee` and `subject`? `email` and `user name` ?

Edit: I see it below in another action but just missing here as well as in the outputs of this action

code42.json
```diff
@@ -1924,6 +1924,463 @@
...
+                    "default": "",
+                    "order": 2
+                },
+                "subject": {
+                    "description": "Subject",
+                    "data_type": "string",
+                    "required": false,
+                    "primary": true,
+                    "contains": [
+                        "email"
```
antazoey: Could also add `user name`

code42.json
```diff
@@ -1924,6 +1924,463 @@
...
+            "render": {
+                "type": "table"
+            },
+            "versions": "EQ(*)"
+        },
+        {
+            "action": "close case",
+            "identifier": "close_case",
+            "description": "Change the status of a Code42 case to 'CLOSED'",
+            "type": "investigate",
```
antazoey, line 2162: I think the `type` here should be something else, probably `contain`.
I basically just make all the write operations `contain` and all the reads `investigate`

code42_connector.py
```diff
@@ -361,6 +362,124 @@ def _handle_remove_legal_hold_custodian(self, param, action_result):
...
+            name,
+            description=description,
+            subject=subject,
+            assignee=assignee,
+            findings=findings,
+        )
+        case_number = response["number"]
+        action_result.add_data(response.data)
+        status_message = f"Case successfully created with case_id: {case_number}"
+        action_result.update_summary(
```
antazoey: Seems like a lot of info for the summary

code42.json
```diff
@@ -1924,6 +1924,463 @@
...
+                    "data_type": "string",
+                    "column_name": "Status",
+                    "column_order": 2
+                },
+                {
+                    "data_path": "action_result.message",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "action_result.summary",
```
antazoey: I think the summary fields should be listed out, like `summary.case_number` etc.
But we are repeating a lot of info in the summary that is already visible in the table and usable in the playbooks and stuff, might be worth just sticking with a case number for a summary and including that in this JSON as such

code42.json
```diff
@@ -1924,6 +1924,463 @@
                 "type": "table"
             },
             "versions": "EQ(*)"
+        },
+        {
+            "action": "create case",
+            "identifier": "create_case",
+            "description": "Create a Code42 case",
+            "type": "investigate",
```
peterbriggs42, line 1932: I haven't seen documentation of what these types mean exactly, but for the other write actions (e.g. `add highrisk employee`, `set alert state`) we've been putting `"type": "contain"`. Do you think that'd be appropriate for `create case` as well?

timabrmsn: Thinking about it a bit, I almost think all of our actions should be in the 'investigate' bucket, since (I think) 'contain' is much more of a response action to a particular event (lock down a device if ransomware detected, etc.), and most Code42 activity fits more into the investigate/observe category.

code42_connector.py
timabrmsn: trimmed these down to just the case number

code42.json
```diff
@@ -1925,6 +1925,505 @@
...
+                {
+                    "data_path": "action_result.data",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "action_result.message",
+                    "data_type": "string"
+                },
+                {
+                    "data_path": "action_result.summary",
```
antazoey: I think should be the case number now

```
"data_path": "action_result.summary.case_number",
"data_type": "numeric"
```

code42.json
```diff
@@ -1925,6 +1925,505 @@
...
+                    "data_path": "action_result.data.*.assigneeUsername",
+                    "data_type": "string",
+                    "contains": [
+                        "email",
+                        "user name"
+                    ],
+                    "column_name": "Assignee",
+                    "column_order": 4
+                },
+                {
```
antazoey, line 2142: Missing `action_result.summary.case_number`

code42.json
```diff
@@ -1925,6 +1925,505 @@
...
+                },
+
+                {
+                    "data_path": "action_result.status",
+                    "data_type": "string",
+                    "column_name": "Status",
+                    "column_order": 2
+                },
+                {
+                    "data_path": "action_result.message",
```
antazoey, line 2423: Missing summary outputs here as well for case num and event id

tests/test_cases_actions.py
```diff
@@ -0,0 +1,309 @@
...
+    mock_py42_client.cases.update.return_value = Py42Response(mock_cases_response)
+    return mock_py42_client
+
+
+@fixture
+def mock_py42_with_cases(mocker, mock_py42_client):
+    response = {"cases": [_TEST_CASE_RESPONSE], "totalCount": 2}
+
+    def gen(*args, **kwargs):
+        yield create_mock_response(mocker, response)
```
antazoey, line 63: Could actually make this yield twice

tests/test_cases_actions.py
```diff
@@ -0,0 +1,309 @@
...
+    mock_py42_client.cases.get_all.side_effect = gen
+    return mock_py42_client
+
+
+class TestCode42CasesConnector(object):
+    def test_handle_action_when_creating_case_calls_with_expected_args_and_sets_success_status(
+        self, mocker, mock_py42_with_case
+    ):
+        mock_user_response = mocker.MagicMock(spec=Response)
+        mock_user_response.text = f'{{"users": [{{"userUid": "{_TEST_USER_UID}"}}]}}'
```
antazoey: We do already have a `mock_py42_with_user` fixture, if that's at all helpful
Your `mock_py42_with_case` fixture could use that as a base fixture instead of  `mock_py42_client`

antazoey (approved):
I think some summaries are missing. This might might be yielding some warnings for you if you try validating.
Looks gooooood though

code42.json
timabrmsn: I don't see any functional difference in the phantom UI when I make this change, or even if I delete this object from the output altogether... Where do these outputs that aren't columns actually show up?

peterbriggs42: @unparalleled-js may know more but I believe that the purpose is for them to be consumed in a playbook context.
https://docs.splunk.com/Documentation/Phantom/4.10.4/PlaybookAPI/Datapaths

peterbriggs42 (approved):

PR #23, opened 2021-06-14
Feature/polling

antazoey:
More testing / parameter work is needed yet.

Here is the just if polling:
1.) Set your polling interval on ingest settings
2.) You can also use "poll now" to force it to check right now
3.) Watch the new events get created from alerts (see screen shot 1)
4.) Notice that the Code42 File Events are also artifacts attached to these Phantom events when applicable (see screen shot 2)

Screen Shot 2021-06-14 at 9 01 06 AM

Screen Shot 2021-06-14 at 9 04 48 AM


logo_code42.svg
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
peterbriggs42: The SVGs in your branch look identical to those in `main`. What's the motivation / nature of this change?

code42_on_poll_connector.py
```diff
@@ -0,0 +1,287 @@
...
+        "video": FileCategory.VIDEO,
+        "virtualdiskimage": FileCategory.VIRTUAL_DISK_IMAGE,
+        "archive": FileCategory.ZIP,
+    }
+    return category_map.get(key, "UNCATEGORIZED")
+
+
+class Code42OnPollConnector:
+    def __init__(self, connector, client, state):
+        super(Code42OnPollConnector, self).__init__()
```
peterbriggs42: Does `Code42OnPollConnector` have a base class defined? If not, what is the purpose of calling `super(...).__init__()` here?

code42_on_poll_connector.py
```diff
@@ -0,0 +1,287 @@
...
+        return container_id
+
+    def _get_container_label(self):
+        return self._connector.get_config().get("ingest", {}).get("container_label")
+
+    def _get_file_events(self, observation, alert_details):
+        query = _get_file_event_query(observation, alert_details)
+        response = self._client.securitydata.search_file_events(query)
+        file_events = response.data.get("fileEvents", [])
+        return [_stringify_lists_if_needed(evt) for evt in file_events]
```
peterbriggs42: Could we call this `e` or `event` rather than `evt`? Seeing `evt` makes my brain think this represents an instance of an acronym that doesn't exist

code42_on_poll_connector.py
```diff
@@ -0,0 +1,287 @@
...
+
+class Code42OnPollConnector:
+    def __init__(self, connector, client, state):
+        super(Code42OnPollConnector, self).__init__()
+        self._connector = connector
+        self._client = client
+        self._state = state or {}
+
+    def handle_on_poll(self, param, action_result):
+        last_time = self._state.get("last_time")
```
peterbriggs42: Is there any Phantom documentation you found for how the polling actions work, or did you just work from the examples of other apps? I'm a bit confused as to how `last_time` gets set on the state object, since you aren't setting it here in the action handler.

code42_on_poll_connector.py
```diff
@@ -0,0 +1,287 @@
...
+            u.get("cloudUsername") for u in shared_with if u.get("cloudUsername")
+        ]
+        event["sharedWith"] = str(shared_list)
+    if private_ip_addresses:
+        event["privateIpAddresses"] = str(private_ip_addresses)
+    return event
+
+
+def _create_artifact_json(container_id, alert_details, file_event):
+    event_data = {
```
peterbriggs42: Nice, great use of a list comprehension here

code42_on_poll_connector.py
```diff
@@ -0,0 +1,287 @@
...
+        key: val for key, val in file_event.items() if val not in [[], None, ""]
+    }
+    return {
+        "container_id": container_id,
+        "source_data_identifier": alert_details["id"],
+        "label": alert_details.get("ruleSource"),
+        "cef": event_data,
```
peterbriggs42: Is our event model in CEF format? I remember seeing in a Slack thread that there *exists* a CEF mapping for file events, but does that imply that the raw event data we receive from the `get_file_events` py42 method already puts the file events into CEF?

logo_code42.svg
antazoey: I accidentally ran an auto formatter that formatted non-py files... I can revert these changes.

code42_on_poll_connector.py
antazoey: Oops, that is no longer needed

code42_on_poll_connector.py
antazoey: I have not worked out `last_time` yet. And there docs for this feature are very much lacking, I am trying to reverse engineer the Jira `on_poll` which does a similar thing

code42_on_poll_connector.py
antazoey: So, it does not really seem like anyone uses the real CEF format here. The most important things is that the data is normalized. Meaning, there is one level and all None values are gone.. Actual CEF fields do not seem to really get used in any of the apps I looked at.  That being said, the doc for this field is `A normalized representation of the data mapped to each field's representative CEF key.`. So it is a tad confusing - In the UI, it looks great and so my thought s were to just do it this way unless someone says we shouldn't

antazoey: Update: CEF is the way to go because of how Phantom playbooks reference the keys from artifacts
It has been updated

unparalleled-js:
A heads up: The CEF format is messed up but I am in the process of correcting it

antazoey:
We can either use CEF (as doing here, borrowed existing mapping from the CLI) or define our own CEF (where C basically stands for "Custom" rather than "Common"). However, we can't share this custom mapping, so if customers wanted to react differently (which is likely) to Code42 alerts, they'd have to know what CEF keys our alert / file-event uses.

I think since we have already have a CEF mapping to just that. The issue there is being familiar with how we map to the CEF keys, which is documented already on our Support site somewhere.

unparalleled-js:
I am trying to add more tests but they are clobbering each other... still trying to figure that out

stubs/phantom/base_connector.py
```diff
@@ -25,10 +33,22 @@ def load_state(self):
...
+        return self._config
 
     def save_state(self, state):
-        pass
+        if not state:
+            return
+        if not self._state:
+            self._state = {**state}
+        else:
+            self._state = {**state, **self._state}
```
timabrmsn: The order of these means the existing state values will override the new values, which isn't what we want, for example:

```
>>> d1 = {"a": 1, "b": 2}
>>> d2 = {"a": 5, "c": 5}
>>> 
>>> {**d1, **d2}
{'a': 5, 'b': 2, 'c': 5}
>>> {**d2, **d1}
{'a': 1, 'c': 5, 'b': 2}
```

We should probably just do `self._state.update(state)`.

code42_on_poll_connector.py
```diff
@@ -0,0 +1,437 @@
...
+        container_label = self._get_container_label()
+        container_json = _create_container(alert_details, container_label)
+        _, _, container_id = self._connector.save_container(container_json)
+        return container_id
+
+    def _get_container_label(self):
+        return self._connector.get_config().get("ingest", {}).get("container_label")
+
+    def _get_file_events(self, param, observations, alert_details):
+        artifact_count = param.get(
```
peterbriggs42, line 146: Does this mean that Phantom imposes a limit on the number of artifacts we can create from a single alert container? If the number of file events associated with the alert exceeds the number of artifact slots, what do we do with the remaining file events?

code42_on_poll_connector.py
```diff
@@ -0,0 +1,437 @@
...
+        "label": container_label,
+    }
+
+
+def _get_file_event_query(observation, alert):
+    mapper = ObservationToSecurityQueryMapper(observation, alert.get("actor"))
+    return mapper.map()
+
+
+class ObservationToSecurityQueryMapper(object):
```
peterbriggs42, line 209: Since we are copy+pasting this from XSOAR, does that suggest that maybe it would be a useful abstraction to add to py42?

code42_on_poll_connector.py
```diff
@@ -0,0 +1,437 @@
...
+    "emailSender": "suser",
+    "eventId": "externalId",
+    "eventTimestamp": "end",
+    "exposure": "reason",
+    "fileCategory": "fileType",
+    "fileName": "fname",
+    "filePath": "filePath",
+    "fileSize": "fsize",
+    "insertionTimestamp": "rt",
+    "md5Checksum": "fileHash",
```
peterbriggs42: Could we also / instead populate the `fileHashMd5` property on the artifact CEF? I'm trying to configure the input to `hunt file` in the playbook to pull the hash from the artifact, but it seems like the `contains` for that parameter only allows the values `fileHashMd5` and `fileHashSha256` from the CEF.

antazoey: Interesting! There seems to be a connection between `contains` and CEF.
Yes, we can just use values from `contains`.

code42_on_poll_connector.py
antazoey: The limit only matters for the `poll now` feature. When you are polling now, you have the ability to provide these parameters. During the regular polling intervals, they are not used.  Also, I set the defaults based on what the UI says, seems like just standard but not necessarily truthful

code42_on_poll_connector.py
antazoey: Yes and no.
My guess is that the alert API is going to improve so that we don't have to do this at all.

peterbriggs42 (approved):

PR #24, opened 2021-06-16
Feature/playbook

peterbriggs42:
Add a playbook that demonstrates the various features of our Phantom integration. 
Add an action `get_user_profile` so that we have access to the user's manager's email address
Update some return values in the JSON so they can be consumed by playbooks
You can use `make playbook` to generate an archive that can be imported into Phantom using the "Import Playbook" button on the Playbooks page.


code42.json
```diff
@@ -1796,6 +1801,176 @@
...
+            "description": "Get user profile",
+            "verbose": "",
+            "type": "generic",
+            "read_only": true,
+            "parameters": {
+                "username": {
+                    "description": "The username associated with the user profile to get",
+                    "data_type": "string",
+                    "contains": [
+                        "email"
```
antazoey: And `user name` ?

code42.json
```diff
@@ -1796,6 +1801,176 @@
...
+                    "order": 0
+                }
+            },
+            "output": [
+                {
+                    "data_path": "action_result.parameter.username",
+                    "data_type": "string",
+                    "column_name": "Username",
+                    "contains": [
+                        "email"
```
antazoey: `user name` here as well

code42.json
```diff
@@ -1796,6 +1801,176 @@
...
+                },
+                {
+                    "column_name": "Manager Username",
+                    "data_path": "action_result.data.*.managerUsername",
+                    "data_type": "string",
+                    "example_values": [
+                        "test@example.com"
+                    ],
+                    "contains": [
+                        "email"
```
antazoey: `user name`

code42.json
```diff
@@ -1796,6 +1801,176 @@
...
+                    ],
+                    "column_order": 8
+                },
+                {
+                    "column_name": "Risk Factors",
```
antazoey: Can we change this `riskFactors` output to be this instead:

```
{
  "column_name": "Risk Tag",
  "data_path": "action_result.data.*.riskFactors.*.tag",
 "column_order": 9,
  "data_type": "string"
},
```

And then in the `code42_connector.py`, in your new action handler, add these lines after getting the response: 

```
all_tags = response.data.get("riskFactors", [])
response["riskFactors"] = _convert_to_obj_list(all_tags, "tag")
```

This makes it usable in Playbooks if needbe

code42_connector.py
```diff
@@ -327,6 +327,13 @@ def _handle_reactivate_user(self, param, action_result):
             phantom.APP_SUCCESS, f"{username} was reactivated"
         )
 
+    @action_handler_for("get_user_profile")
+    def _handle_get_user_profile(self, param, action_result):
+        username = param["username"]
+        profile = self._client.detectionlists.get_user(username)
```
antazoey: If we are in a playbook that calls `create_user` and then after that calls this action, it will fail because the `get_user()` method fails for newly created users for a long time after.

If we use the `get_by_id()` method, it would work better in this case, but that would require having to first get figure out the ID of the user via `user_id = self._get_user_id(username)`

I think it would be worth it though. There are some use-cases involving these two steps

playbook.py
```diff
@@ -0,0 +1,486 @@
...
+from datetime import datetime, timedelta
+def on_start(container):
+    phantom.debug('on_start() called')
+
+    # call 'filter_1' block
+    filter_1(container=container)
+
+    return
+
+def get_alert_details_1(action=None, success=None, container=None, results=None, handle=None, filtered_artifacts=None, filtered_results=None, custom_function=None, **kwargs):
```
antazoey: The styler will love this!

I actually am not sure that we are supposed to be committing this here. Hopefully we can!

playbook.py
```diff
@@ -0,0 +1,486 @@
+"""
```
antazoey: Can we rename this file to `code42_alert_response_playbook.py`

playbooks/Peter Alert Response Playbook.json
```diff
@@ -0,0 +1,7088 @@
+{
```
antazoey: Can we rename this file to `code42_alert_response_playbook.json`?

playbook.py
```diff
@@ -0,0 +1,486 @@
...
+    results_data_1 = phantom.collect2(container=container, datapath=['get_alert_details_1:action_result.data.*.name', 'get_alert_details_1:action_result.parameter.context.artifact_id'], action_results=results)
+    formatted_data_1 = phantom.get_format_data(name='format_1')
+
+    parameters = []
+
+    # build parameters list for 'send_email_1' call
+    for results_item_1 in results_data_1:
+        parameters.append({
+            'from': "Code42 Alert Response",
+            'to': "peter.briggs@code42.com",
```
antazoey: Your email is in here

playbook.py
antazoey: Actually, I think this file is in here twice. This one can be deleted?  And the other one (in the `playbooks/` directory that is named `"Peter ..."`
should have the name above

playbook.py
antazoey: I think it is not in the one in `playbooks/` though, so this file was probably unintentionally left

playbook.py
antazoey: Is there a way we can make `black` ignore this file, since it's autogenerated

playbook.py
peterbriggs42: Yep sorry should have put WIP on the title, there is still some cleanup I intended to do before I submitted this to you and Tim for review

code42.json
peterbriggs42: When I look at the response data for `detectionlists.get_user` the `riskFactors` property looks like a flat string array:
```    
'riskFactors': [   'HIGH_IMPACT_EMPLOYEE',
                       'ELEVATED_ACCESS_PRIVILEGES',
                       'PERFORMANCE_CONCERNS',
                       'FLIGHT_RISK',
                       'SUSPICIOUS_SYSTEM_ACTIVITY'],
```

And when I test the action it shows up in the table without issue. So I think that the current representation in the JSON is ok for playbook integration.

peterbriggs42: Wait nevermind I looked at how `get_high_risk_employee` handles this and I think it makes sense now. Will change

playbook.py
peterbriggs42: Good call, I excluded the `playbooks` directory from the `pre-commit` style checks

playbook.py
peterbriggs42: Yep that playbook file has been removed

code42.json
```diff
@@ -1820,6 +1839,181 @@
...
+                        "test@example.com"
+                    ],
+                    "contains": [
+                        "email",
+                        "user name"
+                    ],
+                    "column_order": 8
+                },
+                {
+                    "column_name": "Risk Factors",
```
antazoey: Can we make the column name `Risk Tags` instead of `Risk Factory` still though?

playbooks/code42_alert_response_playbook.json
```diff
@@ -0,0 +1,7088 @@
...
+                        "approver": "",
+                        "assets": [
+                            {
+                                "action": "create case",
+                                "actions": [
+                                    "run advanced query",
+                                    "run query",
+                                    "hunt file",
+                                    "add case event",
+                                    "list cases",
```
antazoey, line 4379: For some reason, the new action `get user profile` does not appear in this list.
I see it action list assets though.

antazoey (approved):
Found a couple little things

playbooks/code42_alert_response_playbook.json
```diff
@@ -4728,6 +4737,7 @@
                             {
                                 "action": "get user profile",
                                 "actions": [
+                                    "get user profile",
```
antazoey, line 4740: This one is in the list twice

PR #26, opened 2021-06-17
contains & views & polling corrections

antazoey:
* Contains types added
* Custom views aded
The views have titles, cool drop downs for list items (being able to act of each risk tag, for example)
The views have better empty screens
The colors look different (more custom I guess, but just copied the old app and it does not look tooo differnet)
* readme documentation and description updated.
To see the readme, go to Apps -> Code42 and there is a blue link titled "Documentation" near the title  and version of the app.
* Replaces the use of `json.loads(str(_query_))` with `dict(_query_)` everywhere
* Make risk tags primary for add/remove risk tags actions so that it is easier to run from artifacts

Addresses feedback from QA:
* Poll Now also saved state now, the same way as ongoing polling
* We now support the Source Id parameter from the UI (container Id in the code) to limit polled alerts during Poll Now to a specific set of alert IDs
* Fixed bug in alert polling where the checkpoint was not stored as an number. It has to be stored as an int in this setup.
* Fixed bug where alert severities would be override by the default because their artifacts were not given a severity. Now, we pass the severity of the alert onto the artifacts.
* Allows us to set the start and end dates for the first poll in the app config (this is pretty standard)
* Allows us to set the severity of alerts of interest in the app config


code42.json
```diff
@@ -1322,26 +1355,6 @@
                     "column_name": "Description",
                     "column_order": 5
                 },
-                {
```
antazoey, line 1344: I moved parameters to the tops of the outputs list in some cases.

No parameter was harmed in the making of this PR.

readme.html
```diff
@@ -1,10 +1,30 @@
 <html>
   <head></head>
   <body>
-    <h1>Code42 Phantom App</h1>
```
antazoey, line 4: Removed `h1` because the name of the app gets this tag automatically so it looked like we had a duplicated title in the documentation.

readme.html
```diff
@@ -1,10 +1,30 @@
...
-    <p>This app implements various investigative actions on the Code42 Incydr platform.</p>
+    <h2>py42</h2>
```
antazoey, line 4: It is common in phantom to list dependencies first in this file.

It is a great bonus because we get to mention our SDK!

code42_view.py
```diff
@@ -0,0 +1,44 @@
...
+    context["results"] = results = []
+    for summary, action_results in all_app_runs:
+        for result in action_results:
+
+            ctx_result = get_ctx_result(provides, result)
+            if not ctx_result:
+                continue
+            results.append(ctx_result)
+
+    return f"code42_{provides.replace(' ', '_')}.html"
```
antazoey, line 48: This assume all views have the name `code42_<action_id>.html`.

It bothers me how the `provides` argument is the action name with spaces instead of the id with underscores, but that is how it is

code42_search_alerts.html
```diff
@@ -0,0 +1,129 @@
...
+                    <table class="phantom-table dataTable">
+                        <thead>
+                            <th class="widget-th">Actor</th>
+                            <th class="widget-th">State</th>
+                            <th class="widget-th">Severity</th>
+                            <th class="widget-th">Name</th>
+                            <th class="widget-th">Created At</th>
+                            <th class="widget-th">Description</th>
+                        </thead>
+                        <tbody>
```
antazoey, line 74: Allows for which widgets are clickable with contains-based actions and which are just text.

The view that opens up will also give `Indicator` information, which is fascinating. Like if the same user shows up in multiple alerts, etc.

code42.json
```diff
@@ -3150,102 +3295,18 @@
...
-                },
-                {
-                    "data_path": "action_result.parameter.window_title",
-                    "data_type": "string"
                 }
             ],
             "render": {
-                "type": "table"
+                "type": "custom",
+                "view": "code42_view.display_view"
```
antazoey, line 3527: This is what tells the app to use the custom view method from `code42_view.py` `display_view()`

unparalleled-js:
TODO: I noticed there is a little bug in search_alerts. It unnecessarily cuts off precision on the dates.  I will try and fix on this PR (this is like a wrap-up PR)

code42.json
```diff
@@ -75,26 +93,28 @@
                     "allow_list": true
                 },
                 "start_time": {
-                    "data_type": "numeric",
-                    "order": 1,
-                    "description": "Start of time range, in epoch time (milliseconds)",
-                    "verbose": "If not specified, the default is past 10 days"
+                    "description": "The start time to use in the initial poll in %Y-%m-%d %H:%M:%S format (defaults to 30 days back)",
```
antazoey: They are strings since they are input from the configuration page, but what we store / checkpoint is a float.

code42.json
```diff
@@ -1260,7 +1293,10 @@
                         "IN_PROGRESS"
                     ],
                     "default": "",
-                    "order": 1
+                    "order": 1,
+                    "contains": [
+                        "code42 alert state"
```
antazoey, line 1315: The custom contains types should be prefixed with `code42`

playbooks/code42_alert_response_playbook.json
```diff
@@ -4748,7 +4748,6 @@
                                     "create case",
                                     "remove legalhold custodian",
                                     "add legalhold custodian",
-                                    "get user profile",
```
antazoey: @peterbriggs42 this one was in here twice

code42_connector.py
```diff
@@ -333,7 +333,11 @@ def _handle_get_user_profile(self, param, action_result):
         user_id = self._get_user_id(username)
         response = self._client.detectionlists.get_user_by_id(user_id)
         all_tags = response.data.get("riskFactors", [])
+        all_cloud_usernames = response.data.get("cloudUsernames", [])
         response["riskFactors"] = _convert_to_obj_list(all_tags, "tag")
+        response["cloudUsernames"] = _convert_to_obj_list(
```
antazoey, line 342: Also did this for cloud usernames in case some day we add the add/remove cloud usernames actions to this app

CONTRIBUTING.md
```diff
@@ -172,6 +172,22 @@ phenv pip uninstall stubs
...
 
-When we implement actions, describe how to create dummy events in the UI and test the actions via a playbook.
+First, get alerts into Phantom by going to your configured asset's Ingest Settings and enabling
+polling. Then, hit `Poll Now`. You likely will want to increase the `Maximum Containers` and `Maximum Artifacts`
+properties to ingest more alerts.
+
+Go to the Sources tab and look at the newly ingested alerts. Use the "Action" button and select from the action you
+wish to run. Additionally, use the "Playbook" button to run a playbook.
+
+## Reset Polling Checkpoint
```
antazoey, line 185: Might be a good `make` command...
`make clear-checkpoint`

code42.json
```diff
@@ -75,26 +93,28 @@
                     "allow_list": true
                 },
                 "start_time": {
```
antazoey, line 95: Note that the `on poll` parameters _have_ to be these, whereas the app configuration settings can have anything. Thus, apps put the polling-related ingest parameters there instead.

Another really weird thing is that there is no place really in the UI to specify the `start_time` and the `end_time`. It is assumed these will come from another source. In our case, it comes from the app configuration as well as the stored last timestamp ongoingly

code42.json
```diff
@@ -75,26 +93,28 @@
...
+                    "verbose": "Configure this parameter via the `end_date` configuration parameter.",
+                    "data_type": "string",
+                    "required": false,
+                    "order": 2
                 },
                 "container_count": {
                     "data_type": "numeric",
                     "order": 3,
-                    "description": "Maximum number of container records to query for."
+                    "description": "Maximum number of alerts to create (only used in Poll Now)"
```
antazoey, line 112: Only using these params at all because they always show up in the UI no matter what, and I think if it possible for a user to set a value there, we should honor it...

I have seen some apps completely ignore this param anyway though... Same with `container_id`

It does not hurt, imo, to honor it, since it there... Same with the

antazoey: Also: It only shows up in the UI for `Poll Now`, and there is not really a way to configure it for ongoing polls, so I don't use it for ongoing polls at all (it will get all the alerts it finds)

code42.json
```diff
@@ -1,7 +1,7 @@
 {
     "appid": "4d8f53a7-7b12-4d7d-8b01-6575680acf6f",
     "name": "Code42",
-    "description": "This app integrates with Code42 to execute various investigate actions",
+    "description": "Code42 provides simple, fast detection and response to everyday data loss from insider threats by focusing on customer data on endpoints and the cloud.",
```
antazoey, line 4: (copied from xsoar)

code42.json
```diff
@@ -928,7 +938,7 @@
                     "description": "A risk tag to add to the user",
                     "data_type": "string",
                     "required": true,
-                    "primary": false,
+                    "primary": true,
```
antazoey, line 950: This allows us to run `add_highrisk_tag` and `remove_highrisk_tag` directly from the individual tags in the table list from `code42_get_highrisk_employee.html`

code42.json
```diff
@@ -75,26 +93,28 @@
                     "allow_list": true
                 },
                 "start_time": {
-                    "data_type": "numeric",
-                    "order": 1,
-                    "description": "Start of time range, in epoch time (milliseconds)",
-                    "verbose": "If not specified, the default is past 10 days"
+                    "description": "The start time to use in the initial poll in yyyy-MM-dd HH:MM:SS format (defaults to 30 days back)",
+                    "verbose": "Configure this parameter via the `start_date` configuration parameter.",
```
peterbriggs42: Could we specify that this configuration is on the asset? `via the start_date asset configuration parameter` or something like that? 

Though I suppose this text will never be displayed on UI so perhaps it's not necessary...

code42.json
peterbriggs42: Nice!

code42.json
```diff
@@ -3020,16 +3066,125 @@
                     "description": "Names of all the open browser tabs or windows during exposure.",
                     "data_type": "string",
                     "order": 13,
-                    "primary": false
+                    "primary": false,
+                    "contains": [
```
peterbriggs42, line 3079: What is the benefit of adding a custom contains type to parameters like `window title` that are only used on a single action? I could see a use case for this contains if it were an input or output to a different action so you could chain them together, but is it much use as is?

code42_connector.py
```diff
@@ -409,7 +413,8 @@ def _handle_remove_legal_hold_custodian(self, param, action_result):
         if legal_hold_membership_id is None:
             return action_result.set_status(
                 phantom.APP_ERROR,
-                f"Code42: User is not an active member of legal hold matter {matter_id} for action 'remove_legalhold_custodian'.",
+                f"Code42: User is not an active member of "
```
peterbriggs42, line 419: Did `black` do this?

code42_on_poll_connector.py
```diff
@@ -102,8 +102,18 @@ def __init__(self, connector, client, state):
...
         param = self._adjust_date_parameters(param)
-        query = build_alerts_query(param["start_date"], param.get("end_date"))
+        severities = self._connector.get_config().get("severity")
+        if severities:
+            severities = severities.replace(" ", "").split(",")
+        query = build_alerts_query(
+            param["start_time"], param.get("end_time"), severities=severities
+        )
         alerts = self._get_alerts(param, query)
+        source_id = param.get("container_id")
```
peterbriggs42, line 104: Are we sure that this parameter is meant to mean the source ID for the container (aka `source_data_identifier` when we init the container) rather than the event ID that is displayed in the UI?

code42.json
antazoey: It does get displayed in the UI! And yes, I could include that it's the asset

Edit: Done!
https://github.com/phantomcyber/phcode42v2/pull/26/commits/d312746b5f13ef0f43104b9e7bd8f6087493e014

code42.json
antazoey: We can make searches using window title, so we can output one search into another

code42_connector.py
antazoey: No, but black is ok with it and I usually like to do this for longer strings.

code42_on_poll_connector.py
antazoey: This is my understanding:
* The param `container_id` means "only get source events (Code42 alerts, in our case) with these Source IDs (Alert Ids) "
* The alerts are the containers, and the file event-observations of certain alerts become artifacts on the alert
* Artifacts are also a type of container, and both have the property `source_data_identifier` which just mean the external ID (the ID from our system instead of Phantom's).

CONTRIBUTING.md
antazoey: If automated, it will either just have to be a script or use an environment variable to know the asset number to reset rather than `make clear-checkpoint` alone).

code42.json
antazoey: Oh, I see that param is not included in the output for some reason! No worries, I can add it.
I think all the search params for file events should include the data in the output so we can support doing this chain of searches

antazoey: Check it out!  Support for window title chained queries is now a thing:
Screen Shot 2021-06-22 at 08 00 15
Commits:
https://github.com/phantomcyber/phcode42v2/pull/26/commits/5e2353564ce44f82e1c9b65096a6ec1ed8497ae9
and
https://github.com/phantomcyber/phcode42v2/pull/26/commits/359eb4d557b66fe50752395a65a4881c3f9d441d

CONTRIBUTING.md
antazoey: Maybe another PR?

This seems to do the trick:

```
#!/bin/bash

APP_ID="4d8f53a7-7b12-4d7d-8b01-6575680acf6f"

main() {
  asset_number="${1:?Missing parameter 'asset_number' at position 1}"
  # TODO: add `/` in front of opt
  state_file="opt/phantom/local_date/app_states/${APP_ID}/${asset_number}_state.json"

  if [[ -f "${state_file}" ]]; then
    state_json=$(cat $state_file)
    current_value=${state_json:13}
    current_value=${current_value::${#current_value}-1}
    sed -i '' "s/${current_value}/0/g" $state_file
    cat $state_file
  else
    echo "No asset with number ${asset_number}"
  fi
}

main "$@"

```

code42.json
peterbriggs42: Oooh nice this is very cool.

peterbriggs42: Oooh nice this is very cool!

CONTRIBUTING.md
timabrmsn: Can there be an option to reset everything? So you don't need to know the number if there's only one asset configured (or you don't care about multiple being reset)

readme.html
```diff
@@ -1,10 +1,32 @@
...
+    <p>
+      If using the polling feature, you may set the Start Date and End Date for the initial ingest. Otherwise,
+      it ingests up to 30 days back. Ongoing queries will only get new alerts. Configure the polling interval in the
+      Ingest Settings tab. Additionally, you can configure which alert severities to poll for, such as HIGH, MEDIUM,
+      or LOW.
+    </p>
+
+    <h2>On Poll</h2>
+      <p>
+        The 'on poll' functionality first ingests the past 30 days of Code42 alerts (or uses the configured state and
```
timabrmsn: uses the configured *start

code42_on_poll_connector.py
timabrmsn: What's the use-case for polling for a single (or list of) alert ID(s)?

code42.json
timabrmsn: Doesn't `primary` need to be `true` on `alert_state` (and `username` above) for the contains values to be able to be used as inputs? Right now I can't take any actions on the alert state output from a search.

code42_on_poll_connector.py
```diff
@@ -113,8 +123,10 @@ def handle_on_poll(self, param, action_result):
             self._save_artifacts_from_file_events(container_id, details, file_events)
 
         # Save last time of last alert for future polling
-        if not self._connector.is_poll_now() and details.get("createdAt"):
-            self._state["last_time"] = details["createdAt"]
+        last_created_at = details.get("createdAt")
```
peterbriggs42: If you poll once and set checkpoint to, e.g., `2021-06-01` and then you use poll now for a specific set of alerts created on `2021-04-15`, then the checkpoint would be set back to April 15 and next time we auto-poll we'll create duplicate alerts (I think? or maybe best case we'll just fetch them unnecessarily)

Should we only update the checkpoint when `source_id is None`?

code42.json
peterbriggs42: If it's only the `on poll` parameters that have their names fixed to `start_time` / `end_time`, can we change the names of `start_date` and `end_date` to more explicitly call out that they are only used for the initial poll? I keep getting confused about which of these two sets of variables is configured at the asset level vs. action level, and which overrides which.

code42_on_poll_connector.py
antazoey: `start_date` and `end_date` only get used if there is no checkpoint.

antazoey: interesting point about source id though... if you specify a source id and your checkpoint is past that point, you would not get your alerts...

Perhaps they should be exclusive, like you say

code42.json
timabrmsn: I went through the rest of params that have any `contains` values, and the only other one I found missing `primary: true` is the `assignee` field on `list_cases`.

code42_on_poll_connector.py
antazoey: this parameter exists in the UI _no matter what_... Some apps ignore it, but I figured since we can't make it go away, we might as well implement it.

A use case would be something like an analyst knows of a specific alert(s) that is in Code42 but for some reason it did not show up in Phantom.Now, they can manually ingest it.

antazoey: I just made it such that all other query parameters are ignored when specifying source IDs, which really makes the most sense, this was good feedback.

code42_on_poll_connector.py
```diff
@@ -142,25 +164,25 @@ def _init_container(self, alert_details):
     def _get_container_label(self):
         return self._connector.get_config().get("ingest", {}).get("container_label")
 
-    def _get_file_events(self, param, observations, alert_details):
-        artifact_count = param.get(
-            "artifact_count", DEFAULT_ARTIFACT_COUNT_FOR_POLL_NOW
-        )
+    def _get_file_events(self, observations, alert_details, artifact_count=None):
+        def _have_enough_events():
+            do_limit_count = artifact_count is not None
```
antazoey, line 176: Note: Our stylers get mad each other because of PEP `W503 line break before binary operator`...

`black` wanted:

```
return (
                artifact_count is not None
                and is_poll_now
                and len(file_events) >= artifact_count
            )
```

Probably a better way around this, just made the line smaller via a local variable

code42_on_poll_connector.py
timabrmsn: gotcha, totally makes sense to implement something if the field is forced to be there either way.

antazoey:
Seeing errors for subsequent polls now, investigating,
Something complaining an int exceeds int max

code42_on_poll_connector.py
```diff
@@ -101,30 +101,40 @@ def __init__(self, connector, client, state):
...
+                severities = severities.replace(" ", "").split(",")
+            query = build_alerts_query(
+                param["start_time"], param.get("end_time"), severities=severities
+            )
+            alerts = self._get_alerts(param, query)
+            alert_ids = [alert["id"] for alert in alerts]
+            alerts = self._get_alert_details(alert_ids)
 
-        details = {}
+        alert = None
```
timabrmsn: If the `alerts` list is empty, won't this make the `.get()` later raise an exception? Should this be an empty dict instead?

antazoey: Oops, thank you

antazoey: I will add a test for this too
Edit: Done

tests/test_on_poll_connector.py
```diff
@@ -619,3 +619,16 @@ def test_on_poll_when_configured_with_severities_uses_severities(
         assert severity_filters[1]["operator"] == "IS"
         assert severity_filters[1]["term"] == "severity"
         assert severity_filters[1]["value"] == "MEDIUM"
+
+    def test_on_poll_when_no_alerts_is_still_successful(
```
antazoey, line 867: This would fail without the previous commit!

code42_on_poll_connector.py
```diff
@@ -101,30 +101,40 @@ def __init__(self, connector, client, state):
...
+            )
+            severities = self._connector.get_config().get("severity")
+            if severities:
+                severities = severities.replace(" ", "").split(",")
+            query = build_alerts_query(
+                param["start_time"], param.get("end_time"), severities=severities
+            )
+            alerts = self._get_alerts(param, query)
+            alert_ids = [alert["id"] for alert in alerts]
+            alerts = self._get_alert_details(alert_ids)
```
timabrmsn: This will error if the alert_ids list is > 100 too.

antazoey: Right, I forgot about that. Darn!

antazoey: I had forgotten! Just fixed

timabrmsn:
This is looking good to me, the only last thing I'd change is to make these 3 inputs `primary: true` so that the "contains" values work on these actions: 

https://github.com/phantomcyber/phcode42v2/blob/92ba542f0c4a14669049d3a4222f773a0510f8fa/code42.json#L1292
https://github.com/phantomcyber/phcode42v2/blob/92ba542f0c4a14669049d3a4222f773a0510f8fa/code42.json#L1305
https://github.com/phantomcyber/phcode42v2/blob/92ba542f0c4a14669049d3a4222f773a0510f8fa/code42.json#L2634

code42_on_poll_connector.py
```diff
@@ -214,6 +214,7 @@ def _adjust_date_parameters(self, param):
                 "%Y-%m-%dT%H:%M:%S.%f"
             )
             param["start_time"] = last_time_as_date_str
+            param["end_time"] = None
```
unparalleled-js, line 217: This is needed or else something weird gets put here

unparalleled-js:
Oh wow... I just realized I am saving the newest timestamp instead of the oldest...
How was this working before...

code42_on_poll_connector.py
```diff
@@ -196,27 +201,32 @@ def _save_artifacts_from_file_events(self, container_id, details, file_events):
         ]
         self._connector.save_artifacts(artifacts)
 
-    def _adjust_date_parameters(self, param):
+    def _adjust_date_parameters(self):
```
peterbriggs42, line 204: This is a lot easier to read / understand now. Nice work!

peterbriggs42 (approved):

code42_util.py
```diff
@@ -40,4 +40,5 @@ def build_alerts_query(
         filters.append(Severity.is_in(severities))
     filters.append(build_date_range_filter(DateObserved, start_date, end_date))
     query = AlertQuery.all(*filters)
+    query.sort_direction = "asc"
```
antazoey, line 43: @peterbriggs42 @timabrmsn  is this change okay with both of you??
We should be getting the earlier polls first, it makes more sense
(polling was not working before because of this + that we get the last alert timestamp)

timabrmsn: yep, makes sense!

timabrmsn (approved):

PR #27, opened 2021-06-21
Cases testing feedback

timabrmsn:
- `list_cases` missed converting username -> userUid, so this update fixes that. 
- Adds better error messages to 404s (and 400s for what should be 404s on `add_case_event`)

### DIFF
--- a/code42_connector.py
+++ b/code42_connector.py
@@ -11,6 +11,7 @@ import requests
 from phantom.action_result import ActionResult
 from phantom.base_connector import BaseConnector
 from phantom.vault import Vault
+from py42.exceptions import Py42BadRequestError
 from py42.exceptions import Py42NotFoundError
 from py42.exceptions import Py42UpdateClosedCaseError
 from py42.sdk.queries.fileevents.file_event_query import FileEventQuery
@@ -405,6 +406,8 @@ class Code42Connector(BaseConnector):
         status_message = f"{username} was removed from legal hold matter {matter_id}."
         return action_result.set_status(phantom.APP_SUCCESS, status_message)
 
+    """ CASES ACTIONS """
+
     @action_handler_for("create_case")
     def _handle_create_case(self, param, action_result):
         name = param["case_name"]
@@ -441,14 +444,17 @@ class Code42Connector(BaseConnector):
             assignee = self._get_user_id(assignee)
         description = param.get("description")
         findings = param.get("findings")
-        response = self._client.cases.update(
-            case_number,
-            name=name,
-            subject=subject,
-            assignee=assignee,
-            description=description,
-            findings=findings,
-        )
+        try:
+            response = self._client.cases.update(
+                case_number,
+                name=name,
+                subject=subject,
+                assignee=assignee,
+                description=description,
+                findings=findings,
+            )
+        except Py42NotFoundError as err:
+            self._handle_case_not_found(err, case_number)
         status_message = f"Case number {case_number} successfully updated"
         action_result.add_data(response.data)
         action_result.update_summary({"case_number": case_number})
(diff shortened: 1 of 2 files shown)


code42_connector.py
```diff
@@ -491,11 +503,30 @@ def _handle_list_cases(self, param, action_result):
...
+            )
+        except Py42BadRequestError as err:
+            if "NO_SUCH_CASE" in err.response.text:
+                self._handle_case_not_found(err, case_number)
+            elif "NO_SUCH_EVENT" in err.response.text:
+                message = f"Event ID {event_id} not found."
+                err.args = (message,)
+                raise err
+            else:
+                raise err
```
antazoey: This should just be `raise` instead of `raise err` to preserve the exception stack

code42_connector.py
```diff
@@ -491,11 +503,30 @@ def _handle_list_cases(self, param, action_result):
     def _handle_add_case_event(self, param, action_result):
         case_number = param["case_number"]
         event_id = param["event_id"]
-        self._client.cases.file_events.add(case_number=case_number, event_id=event_id)
+        try:
+            self._client.cases.file_events.add(
+                case_number=case_number, event_id=event_id
+            )
+        except Py42BadRequestError as err:
+            if "NO_SUCH_CASE" in err.response.text:
```
antazoey, line 511: I wish these were just in py42 instead

tests/test_cases_actions.py
```diff
@@ -247,7 +247,7 @@ def test_handle_action_when_list_cases_calls_py42_with_correct_filters(
         connector = create_fake_connector("list_cases", mock_py42_with_case)
         connector.handle_action(param)
         mock_py42_with_case.cases.get_all.assert_called_once_with(
-            status="OPEN", assignee=_TEST_ASSIGNEE, subject=_TEST_SUBJECT
+            status="OPEN", assignee=_TEST_USER_UID, subject=_TEST_USER_UID
```
antazoey, line 250: If we wanted to get fancier here, we could change the `get_user` side effect to return different IDs based on if it received `_TESET_ASSIGNEE`, or `_TEST_SUBJECT` (and just use `_TEST_USER_UID` if neither, if that is needed)... But by doing this would be accurately testing this feature, otherwise there is a bit of gap.

code42_connector.py
```diff
@@ -473,7 +481,11 @@ def _handle_list_cases(self, param, action_result):
         if status == "ALL":
             status = None
         assignee = param.get("assignee")
+        if assignee is not None:
```
antazoey: Probably safer to just do `if assignee` so that we count `""` as `None`

antazoey: Same with `if subject`

code42_connector.py
timabrmsn: Yeah, ideally we'd add a `Py42CaseNotFoundError` and maybe a `Py42EventNotFoundError` too, this would clean up the CLI errors for missing cases too.

antazoey (approved):
Improving the test for assignee / subject seems like an easy win still

tests/test_cases_actions.py
timabrmsn: I think the main thing we care about is that `get_all()` is called with the userUid instead of the passed usernames for both of those params, so it doesn't really matter that they're different, just that the value comes from the `get_by_username()` calls.

timabrmsn: I suppose we could also test that `get_by_username()` was called twice since there are two params that need it called.

antazoey: Checking the call count _is_ better, but theoretically it would still pass if we called `_get_user()` for the assignee twice (instead of 1 each)

PR #29, opened 2021-06-23
Upgrade py42

antazoey:
I created the py42 wheel via the command 
```bash
python setup.py bdist_wheel
```

While on the main branch, fully caught up. It output `/wheels/py42-1.15.1-py2.py3-none-any.whl`.

I placed it into the phantom app and changed all its reference. Then, I deployed it and verified I can still connect to Code42 using py42.


tests/test_util.py
```diff
@@ -26,9 +26,23 @@ def test_build_alerts_query_returns_expected_query():
     _assert_on_or_before(actual["groups"][0]["filters"][1], end_date)
 
 
-def test_build_alerts_query_when_given_username_returns_expected_query():
+def test_build_alerts_query_when_using_milliseconds_returns_expected_query():
```
antazoey, line 29: Also added an explicit milliseconds test just in case it would be needed

timabrmsn (approved):

PR #31, opened 2021-06-23
Pollnow fix

antazoey:
Poll Now is supposed to use the last time as well.
Can't I believe I missed this!
QA pointed it out.

Also added a couple other tests as I was investigating the bug, related to `artifact_count` exceeding elements in a list. Was no bug there but tests are nice.

### DIFF
--- a/tests/test_on_poll_connector.py
+++ b/tests/test_on_poll_connector.py
@@ -580,7 +580,7 @@ class TestCode42OnPollConnector(object):
         connector.handle_action(param)
         assert_artifacts_added(connector, [EXPECTED_ARTIFACTS[0]])
 
-    def test_on_poll_when_is_poll_and_artifact_count_exceeds_event_count_creates_artifacts_for_all(
+    def test_on_poll_when_is_poll_now_and_artifact_count_exceeds_event_count_creates_artifacts_for_all(
         self, mock_py42_for_alert_polling
     ):
         connector = _create_on_poll_connector(mock_py42_for_alert_polling)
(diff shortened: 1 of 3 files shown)


tests/test_on_poll_connector.py
```diff
@@ -665,6 +684,23 @@ def get_alert_details(alert_id, *args, **kwargs):
         assert connector._state is None
         assert_success(connector)
 
+    def test_on_poll_when_is_poll_now_uses_last_time(
```
unparalleled-js, line 687: test that addresses bug

tests/test_on_poll_connector.py
```diff
@@ -605,6 +624,26 @@ def test_on_poll_when_is_not_poll_now_uses_previously_stored_timestamp(
         assert abs((actual_date - expected_date)).seconds < 1
         assert_success(connector)
 
+    def test_on_poll_when_is_poll_now_uses_previously_stored_timestamp(
```
antazoey, line 629: This test tests the fix

the rest are bonuses

tests/test_on_poll_connector.py
```diff
@@ -665,6 +738,23 @@ def get_alert_details(alert_id, *args, **kwargs):
         assert connector._state is None
         assert_success(connector)
 
+    def test_on_poll_when_is_poll_now_and_specifying_source_ids_does_not_store_last_time(
```
antazoey, line 741: just making sure things work as normal with poll_now, since bugs

peterbriggs42 (approved):

PR #34, opened 2021-08-03
Correctly handle timezones

antazoey:
* Related to the bug that QA found in splunk, only fixing it in Phantom.
More information:
https://blog.ganssle.io/articles/2019/11/utcnow.html


alanag13:
it looks like gh actions no longer supports python 3.6.13, which is unfortunate -- hopefully bumping the the version to .14 should allow everything to still work

alanag13 (approved):

